PR #12, opened 2021-08-03
refactor: remove spurious scripts

### LINKED ISSUE #4: Remove `<script>` references to CDN
To limit the number of source locations that AskGov needs to run, all external `<script>` entries in index.html should be examined and removed or replaced with equivalent npm entries, with corresponding changes to the frontend. Relevant lines in Content Security Policy are also to be removed.

Lines to be removed include:
https://github.com/opengovsg/askgovsg/blob/85bfcc2b7720bb9bdfa6934ef8060dc51245bed8/client/public/index.html#L23-L38

https://github.com/opengovsg/askgovsg/blob/85bfcc2b7720bb9bdfa6934ef8060dc51245bed8/server/src/bootstrap/helmet-options.ts#L26-L31

https://github.com/opengovsg/askgovsg/blob/85bfcc2b7720bb9bdfa6934ef8060dc51245bed8/server/src/bootstrap/helmet-options.ts#L38-L39

nauynix:
## Problem

There are four scripts called in `index.html` for CDN. External <script> entries in index.html should be examined and removed or replaced with equivalent npm entries

Closes #4 

## Solution

The four packages `stacks-icons`, `jquery`, `popper`, `bootstrap` are not used in our codebase. They were put there by the initial author of `stackoverflow-clone`. Since we are using react, and box-icons for our icons, we do not use the scripts at all. The relevant permissions in Content Security Policy are also removed.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -7,7 +7,6 @@
     "@emotion/react": "^11.4.0",
     "@emotion/styled": "^11.3.0",
     "@fullstory/browser": "^1.4.9",
-    "@stackoverflow/stacks-icons": "2.20.0",
     "@styled-icons/boxicons-logos": "^10.23.0",
     "@styled-icons/boxicons-regular": "^10.34.0",
     "@testing-library/jest-dom": "^4.2.4",
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -3240,11 +3240,6 @@
         "@sinonjs/commons": "^1.7.0"
       }
     },
-    "@stackoverflow/stacks-icons": {
-      "version": "2.20.0",
-      "resolved": "https://registry.npmjs.org/@stackoverflow/stacks-icons/-/stacks-icons-2.20.0.tgz",
-      "integrity": "sha512-IxRt8SCSXmOAnvLNd4vIb2xkflNwuksfBBinLUSO0xv8mhGq5IKjiCnnSd4Iu59dnrOxUQVjzGeJcMwDNbctZg=="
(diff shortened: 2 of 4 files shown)


nauynix:
While care has been taken to comb through the codebase to check that the packages are indeed, not used, this is stil lthe first time I've encounted `jquery` and `bootstrap`. Please help to double-check that they are really not inside the codebase.

client/package.json
```diff
@@ -7,7 +7,6 @@
     "@emotion/react": "^11.4.0",
     "@emotion/styled": "^11.3.0",
     "@fullstory/browser": "^1.4.9",
-    "@stackoverflow/stacks-icons": "2.20.0",
     "@styled-icons/boxicons-logos": "^10.23.0",
     "@styled-icons/boxicons-regular": "^10.34.0",
```
LoneRifle, line 11: I see no reference to this in the codebase. Safe to remove?

nauynix: logos is used in footer, regular is used in editbutton

LoneRifle:
Holding off on this for now until after release

client/package.json
nauynix: we should probably get all of the icon packages from one source, but I don't know which source to use. Right now its 3/4 different one with no clear preferred way.

LoneRifle: we should use `react-icons/bi`, I would imagine.

LoneRifle (approved):
lgtm

PR #25, opened 2021-08-04
test(mail): add unit tests, supporting cast

LoneRifle:
## Problem

With impending features requested for the next iteration of AskGov, a test framework should be put in place so that
the team is encouraged to provide unit tests along with their implementations. This PR serves to do that, using
MailService as an example.

To facilitate unit-testing of MailService, we also need to do some light refactoring.

## Solution

Change the ctor so that we accept a subset of nodemailer's Transporter. 
This allows us to more easily supply a mock for testing.

- change `mailOptions: SMTPTransport.Options` to
  `transport: Pick<Transporter, 'sendMail'>`
- rework bootstrap to inject an actual nodemailer Transporter

- provide a test for MailService
- include npm run scripts to facilitate testing
  - invoke client tests with `--passWithNoTests` for now since there
    are none
- tighten `server/tsconfig.json` so that it does not pick up files outside the specified `rootDir: src/`
- reinstate test step in GitHub Actions

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "postinstall": "npm --prefix server install && npm --prefix client install",
     "auditDep": "npm audit fix && npm --prefix server audit fix && npm --client server audit fix",
     "prepare": "husky install",
+    "test": "npm --prefix server test && CI=true npm --prefix client test -- --passWithNoTests",
     "lint": "npm run --prefix server lint && npm run --prefix client lint",
     "lint-ci": "npm run --prefix server lint-ci && npm run --prefix client lint-ci"
   },
--- a/server/src/modules/mail/mail.service.ts
+++ b/server/src/modules/mail/mail.service.ts
@@ -1,18 +1,17 @@
-import { createTransport, Transporter } from 'nodemailer'
-import SMTPTransport from 'nodemailer/lib/smtp-transport'
+import { Transporter } from 'nodemailer'
 import { renderLoginOtpBody } from './mail.util'
 
 export class MailService {
   private mailFromEmail: string
-  private transport: Transporter
+  private transport: Pick<Transporter, 'sendMail'>
   constructor({
-    mailOptions,
+    transport,
     mailFromEmail,
   }: {
-    mailOptions: SMTPTransport.Options
+    transport: Pick<Transporter, 'sendMail'>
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -14,6 +14,7 @@
     "postinstall": "npm --prefix server install && npm --prefix client install",
     "auditDep": "npm audit fix && npm --prefix server audit fix && npm --client server audit fix",
     "prepare": "husky install",
+    "test": "npm --prefix server test && CI=true npm --prefix client test -- --passWithNoTests",
```
nauynix, line 17: Can I check if the -- before --passWithNoTests is required? What does it do?

server/src/modules/mail/mail.service.ts
```diff
@@ -1,18 +1,17 @@
-import { createTransport, Transporter } from 'nodemailer'
-import SMTPTransport from 'nodemailer/lib/smtp-transport'
+import { Transporter } from 'nodemailer'
 import { renderLoginOtpBody } from './mail.util'
 
 export class MailService {
   private mailFromEmail: string
-  private transport: Transporter
+  private transport: Pick<Transporter, 'sendMail'>
   constructor({
```
nauynix, line 7: Theres syntactic sugar for this, credit to @mantariksh.
https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties

server/src/modules/mail/mail.service.ts
```diff
@@ -1,18 +1,17 @@
...
     mailFromEmail,
   }: {
-    mailOptions: SMTPTransport.Options
+    transport: Pick<Transporter, 'sendMail'>
     mailFromEmail: string
   }) {
-    this.transport = createTransport(mailOptions)
+    this.transport = transport
     this.mailFromEmail = mailFromEmail
   }
```
nauynix, line 16: Theres syntactic sugar for this if you want. Credits to @mantariksh 
https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties

package.json
LoneRifle: `--` passes the arguments after it to the command invoked by the npm run script, so it's needed

nauynix (approved):
lgtm

PR #29, opened 2021-08-04
build(deps): audit and update

LoneRifle:
## Problem and Solution

Inspect and update dependencies that npm highlights as audit issues


barnettx (approved):
lgtm, was this done with `npm audit fix`?

LoneRifle:
Indeed it is!

Issue #17, opened 2021-08-03
User can fill a form to ask questions to agency


LoneRifle:
Currently, we use an Email Us with `mailto:`. Changing this to a form where a user fills up gives us the following benefits:

- As a user is typing the question, we can show other already answered questions (like StackOverflow), so that user can refer to those and get their answers faster.
- We can also store what questions users are asking for further analysis (e.g. are these new questions that we should get agencies to create FAQs for, or existing questions such that users are not searching)

LoneRifle:
> Changing this to a form where a user fills up gives us the following benefits
> 
> * As a user is typing the question, we can show other already answered questions (like StackOverflow), so that user can refer to those and get their answers faster.

To my mind, this veers towards Phase 3 and can be counted as scope creep.

We can perhaps reuse the search components to anticipate what the user might be asking for, but we would still need to build out the backend to route the submitted request to the agency. 

Alternatively, the form could just generate the corresponding e-mail to send to the agency + ourselves, but this would oblige the user to go through an extra step to send the e-mail, which isn't good UX

> * We can also store what questions users are asking for further analysis (e.g. are these new questions that we should get agencies to create FAQs for, or existing questions such that users are not searching)

As an alternative to storing the questions, we could have ourselves cc'd in the mail sent to agencies with the appropriate parameters added to the `mailto:` link.

### Proposed iterations

I hence propose the following iterations to get to the functionality we want:

- ~We will initially add a cc to ourselves for the `mailto:` link generated for the agency.~ done
- A second iteration will present a form to the user to fill up a form that would generate the mail as described above
- Subsequently, we build out the backend needed to:
  - Initially route the form submission into e-mails, which we then replace with... 
  - ... Hold the request as a question to be answered in the application itself

Issue #23, opened 2021-08-03
Clicking on topic tag after navigating from agency page doesn't scope to agency


LoneRifle:
When navigating from agency page to post detail page, clicking on topic tag will scope to show posts with that topic tag AND agency tag eg. visit post of 'Am I entitled to a 10-minute grace period after I have parked via the app?' Clicking `start-session` should navigate to `/agency/parkingsg?tags=start-session`, including `parkingsg` in the scope. As it is, it navigates to `?tags=start-session`

## Screenshots
**Note the URL**
**BEFORE**

https://user-images.githubusercontent.com/18087283/124701850-d8d9cf80-df21-11eb-818d-2485a104f99c.mov

**AFTER**

https://user-images.githubusercontent.com/18087283/124701792-bd6ec480-df21-11eb-97f5-80c605f4b265.mov

LinHuiqing:
Closed with https://github.com/opengovsg/askgovsg/issues/491?

LoneRifle:
problem might still be present, unsure

PR #31, opened 2021-08-04
feat: create new email modal component

nauynix:
## Problem
Currently, the enquiry button leads to a mailto:.

- Addresses the second iteration of #17

## Solution

This PR adds a new enquiry modal, which the enquiry button will open up to. To reduce the scope of changes, the submission of the modal will still lead to a prepopulated `mailto:`. Automatic mailing of the enquiry will be another PR. The `Sender Email` design code is added and commented out in anticipation of the next PR.

There are two places where the user can open up the enquiry modal.
In the homepage, it will open up to the specific agency if it is selected, and default to `feedback@ask.gov.sg`.
In the post page, it will always open up the first agency that was tagged with the question. This is because the back button also does the same thing. It might be better to grab all the agencies tagged with it instead.

## Before & After Screenshots

**BEFORE**:

https://user-images.githubusercontent.com/20250559/128158436-9825e0c7-dcd1-46a1-abb0-b51674a975ce.mov

**AFTER**:

https://user-images.githubusercontent.com/20250559/128158242-99b76d64-c000-4651-b953-3a5259b53f20.mov

## Tests
- [x] Modal does not accept empty form
- [x] Modal will clear upon submission
- [x] Modal confirm leads to `mailto:` with address and CC appropriately filled out.

### DIFF
--- a/client/src/components/CitizenRequest/CitizenRequest.component.tsx
+++ b/client/src/components/CitizenRequest/CitizenRequest.component.tsx
@@ -10,7 +10,7 @@ export interface Enquiry {
 }
 
 const CitizenRequest = ({ agency }: { agency: Agency }): JSX.Element => {
-  const cc = agency.email === 'feedback@ask.gov.sg' ? '' : 'feedback@ask.gov.sg'
+  const cc = agency.email === 'enquires@ask.gov.sg' ? '' : 'enquires@ask.gov.sg'
   const {
     onOpen: onDeleteModalOpen,
     onClose: onDeleteModalClose,
--- a/client/src/components/CitizenRequest/CitizenRequest.component.tsx
+++ b/client/src/components/CitizenRequest/CitizenRequest.component.tsx
@@ -2,10 +2,6 @@ import { Text, Button, Flex, useDisclosure } from '@chakra-ui/react'
 import { EnquiryModal } from '../EnquiryModal/EnquiryModal.component'
 import { Agency } from '../../services/AgencyService'
 
-interface CitizenRequestProps {
-  agency?: Agency
-}
-
 // TODO: combine interface Enquiry from both client and server
 export interface Enquiry {
   questionTitle: string
@@ -13,22 +9,13 @@ export interface Enquiry {
   senderEmail: string
 }
 
-const CitizenRequest = ({
-  agency = {
-    id: '',
-    email: 'askgov@open.gov.sg',
-    shortname: 'AskGov',
-    longname: 'AskGov',
-    logo: '',
-  },
-}: CitizenRequestProps): JSX.Element => {
+const CitizenRequest = ({ agency }: { agency: Agency }): JSX.Element => {
   const cc = agency.email === 'askgov@open.gov.sg' ? '' : 'askgov@open.gov.sg'
   const {
     onOpen: onDeleteModalOpen,
     onClose: onDeleteModalClose,
     isOpen: isDeleteModalOpen,
   } = useDisclosure()
-
   const onPostConfirm = async (enquiry: Enquiry): Promise<void> => {
     window.location.href =
       'mailto:' +
@@ -60,7 +47,6 @@ const CitizenRequest = ({
           background: 'primary.600',
         }}
         borderRadius="4px"
-        // colorScheme="blue"
         color="white"
         onClick={onDeleteModalOpen}
       >
--- a/client/src/components/CitizenRequest/CitizenRequest.component.tsx
+++ b/client/src/components/CitizenRequest/CitizenRequest.component.tsx
@@ -1,16 +1,46 @@
-import React from 'react'
-import { Text, Button, Flex } from '@chakra-ui/react'
+import { Text, Button, Flex, useDisclosure } from '@chakra-ui/react'
+import { EnquiryModal } from '../EnquiryModal/EnquiryModal.component'
+import { Agency } from '../../services/AgencyService'
 
 interface CitizenRequestProps {
-  email?: string
-  longName?: string
+  agency?: Agency
+}
+
+// TODO: combine interface Enquiry from both client and server
+export interface Enquiry {
+  questionTitle: string
+  description: string
+  senderEmail: string
 }
 
 const CitizenRequest = ({
-  email = 'askgov@open.gov.sg',
-  longName = '',
+  agency = {
+    id: '',
+    email: 'askgov@open.gov.sg',
+    shortname: 'AskGov',
+    longname: 'AskGov',
+    logo: '',
+  },
 }: CitizenRequestProps): JSX.Element => {
-  const cc = email === 'askgov@open.gov.sg' ? '' : 'askgov@open.gov.sg'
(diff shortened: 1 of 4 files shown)


client/src/components/EmailModal/EmailModal.component.tsx
```diff
@@ -0,0 +1,146 @@
...
+              {`${agency.shortname.toUpperCase()} Enquiry Form`}
+            </Text>
+          </ModalHeader>
+          <ModalCloseButton />
+          <ModalBody>
+            <VStack align="left" spacing={0}>
+              <Text>
+                {`This enquiry form will be sent to ${agency.longname}. Please do
+                note that we would take at least 3 working days to process your
+                enquiry. Thank you.`}
```
LoneRifle: Some indication here that this will generate an e-mail as an extra step would be helpful to the user

client/src/components/EmailModal/EmailModal.component.tsx
```diff
@@ -0,0 +1,146 @@
...
+                errorBorderColor="error.500"
+                isInvalid={formErrors.senderEmail}
+                placeholder="jony_tan@gmail.com"
+                {...register('senderEmail', {
+                  required: true,
+                  pattern: /^\S+@\S+$/i,
+                })}
+              />
+              {formErrors.senderEmail &&
+                errorLabel('Please enter a valid email')} */}
```
LoneRifle: Is this being put in place ahead of time?

client/src/components/EmailModal/EmailModal.component.tsx
```diff
@@ -0,0 +1,146 @@
...
+              </Text>
+              <Text textStyle="body-2" color="secondary.400">
+                Please fill in a valid email so that we can get back to you
+              </Text>
+              <Box h={3} />
+              <Input
+                focusBorderColor="primary.500"
+                errorBorderColor="error.500"
+                isInvalid={formErrors.senderEmail}
+                placeholder="jony_tan@gmail.com"
```
LoneRifle: Avoid having an e-mail address from an actual domain here. There is no apparent consensus as to what should be an appropriate placeholder for e-mail fields; will leave it at your discretion to have something sensible.

@khaleedah in case she wants to weigh in

client/src/components/CitizenRequest/CitizenRequest.component.tsx
```diff
@@ -1,16 +1,45 @@
...
 const CitizenRequest = ({
-  email = 'askgov@open.gov.sg',
-  longName = '',
+  agency = {
+    id: '',
+    email: 'askgov@open.gov.sg',
+    shortname: 'AskGov',
+    longname: 'AskGov',
+    logo: '',
+  },
```
LoneRifle: This may come back to haunt us, since it will make it awkward to reason about agency later on, especially since right now, we derive this from the presence of agency information in the URL. I would suggest not having a default here, and instead consider the following:

- HomePage - if not bound to an agency (ie, if rendering the government-wide page), should we allow enquiries to be made at all? If so, what would be an appropriate choice of agency to inject into CitizenRequest?
- Post - related to #23 - if navigating from the government-wide page, can we perhaps infer the agency from the Post agency tags?

@mantariksh - significant proposal made by me, in case I have blind spots.

client/src/components/CitizenRequest/CitizenRequest.component.tsx
```diff
@@ -32,20 +61,16 @@ const CitizenRequest = ({
...
-            cc +
-            '&subject=' +
-            longName +
-            ' AskGov&body=Let%20us%20know%20how%20we%20can%20help%20below!'
-        }}
+        onClick={onDeleteModalOpen}
       >
         Submit an enquiry
       </Button>
+      <EmailModal
```
LoneRifle: `EnquiryModal`, perhaps?

LoneRifle (changes requested):
Kinda glad we split this effort in two - there's a bit to unpack here.

LoneRifle:
Also - if you could - rebase your changes so that you can pick up the CI tests

client/src/components/EmailModal/EmailModal.component.tsx
nauynix: @khaleedah can you look into how to reword this?

client/src/components/EmailModal/EmailModal.component.tsx
nauynix: Yes, since the email is currently sent from the user, this field is not needed. Its for the next iteration when we send from askgov but loop the user in

client/src/components/CitizenRequest/CitizenRequest.component.tsx
nauynix: My two cents are that users should have a centralised enquiry place if they do not know which agency to ask, since that is part of what AskGov hopes to aid with.

client/src/components/EnquiryModal/EnquiryModal.component.tsx
```diff
@@ -0,0 +1,146 @@
...
+              {`${agency.shortname.toUpperCase()} Enquiry Form`}
+            </Text>
+          </ModalHeader>
+          <ModalCloseButton />
+          <ModalBody>
+            <VStack align="left" spacing={0}>
+              <Text>
+                {`This enquiry form will be sent to ${agency.longname}. Please do
+                note that we would take at least 3 working days to process your
+                enquiry. Thank you.`}
```
LoneRifle: How about...
```suggestion
                {`This enquiry form will generate an email to be sent to 
                ${agency.longname}. Please note that we would take at 
                least 3 working days to process your enquiry. Thank you.`}
```

client/src/components/EmailModal/EmailModal.component.tsx
nauynix: changing to email

client/src/pages/HomePage/HomePage.component.jsx
```diff
@@ -223,9 +223,17 @@ const HomePage = ({ match }) => {
...
       <CitizenRequest
-        className="citizen-request"
-        email={agency?.email}
-        longName={agency?.longname}
+        agency={
+          agency
+            ? agency
+            : {
+                id: '',
+                email: 'enquires@ask.gov.sg',
```
LoneRifle:
```suggestion
                email: 'enquiries@ask.gov.sg',
```

client/src/pages/Post/Post.component.jsx
```diff
@@ -123,9 +123,17 @@ const Post = () => {
...
       <CitizenRequest
-        className="citizen-request"
-        email={agency?.email}
-        longName={agency?.longname}
+        agency={
+          agency
+            ? agency
+            : {
+                id: '',
+                email: 'enquires@ask.gov.sg',
```
LoneRifle:
```suggestion
                email: 'enquiries@ask.gov.sg',
```

LoneRifle (approved):
lgtm for now. Not sure about the duplicate AskGov agency default, but I assume that this will change when the form does a POST instead of generating an email.

PR #40, opened 2021-08-06
build(deps): use react-icons only, prune unused deps

### LINKED ISSUE #24: Use `react-icons/bi` for box icons
we should probably get all of the icon packages from one source. Right now its 3/4 different one with no clear preferred way.

_Originally posted by @nauynix in https://github.com/opengovsg/askgovsg/pull/12#discussion_r681454422_

LoneRifle:
## Problem and Solution

To keep our dependencies to a minimum, replace all icon occurrences with react-icons,
and prune all unused dependencies

- Remove react-logger, reselect, add @typescript-eslint
- Replace all occurrences of box-icon with react-icon equivalents
  - move margin CSS into the element itself
  - use chakra color codes
- Replace LinkedIn and Facebook icons with react-icon equivalents
  - spacing between social media icons has changed as a result, but
    remains usable
- Remove now-unused deps


client/src/pages/Post/Post.component.jsx
```diff
@@ -137,7 +139,7 @@ const BackButtonForPostDetail = ({ agencyShortName }) => {
   return (
     <div className="back-to-home">
       <button onClick={() => history.push(`/agency/${agencyShortName}`)}>
-        <box-icon size="sm" color="#69738E" name="arrow-back"></box-icon>
+        <BiArrowBack size="24" color="secondary.400" />
```
nauynix:
```suggestion
        <BiArrowBack
          style={{ marginRight: '8px' }}
          size="24"
          color="secondary.400"
        />
```

nauynix: Same margin as the normal back button in `backtohome`

nauynix (changes requested):
lgtm

nauynix (approved):

nauynix:
When I try to run the branch, it fails to compile as it cannot find the package `react-router`. This does not happen with the develop branch, not sure why

https://github.com/opengovsg/askgovsg/blob/49873ba8e44e097b326f2a091aae2203f4e11c83/client/src/components/Header/Header.component.jsx#L3

client/package-lock.json
```diff
@@ -5487,34 +5519,6 @@
...
-            "history": "^4.7.2",
-            "invariant": "^2.2.4",
-            "loose-envify": "^1.3.1",
-            "prop-types": "^15.6.1",
-            "react-router": "^4.3.1",
-            "warning": "^4.0.1"
-          }
-        }
-      }
-    },
```
nauynix, line 5517: Update: `boxicons` requires `react-router` which was why it was imported inside. Since we are removing `boxicons`, all uses of `react-router` should be `react-router-dom`.

LoneRifle: That's already the case, no?

PR #41, opened 2021-08-06
build(ci): add dependabot and mergify to manage deps

LoneRifle:
## Problem

It is difficult to keep up with dependency updates

## Solution

Employ dependabot and mergify to auto-merge minor dependency updates


nauynix (approved):
lgtm

PR #42, opened 2021-08-06
build(deps): replace node-sass with sass

LoneRifle:
## Problem

node-sass is deprecated and has several potential security issues as reported by Snyk

## Solution

replace node-sass with sass


nauynix (approved):
lgtm

PR #43, opened 2021-08-09
Feat: automate sending enquiry as email

### LINKED ISSUE #17: User can fill a form to ask questions to agency

nauynix:
## Problem
Currently, the enquiry button leads to a mailto:. It should instead automatically send a email so the user does not have to click an extra button. 

- Addresses the third iteration of #17

## Solution

Email is sent using the same Amazon Simple Email Service that sends the One Time Password. It will set the `replyTo` to sender email so that agencies can reply to the user. It also sets bcc to sender email rather than `to` or `cc` so that user gets a copy of the email but reduces clutter in the headers. Finally it ccs `enquiries@ask.gov.sg` so we keep a copy of the email and stay in the loop.
- Expose a new endpoint `/enquiries` that accepts an array `agencyID` and details of enquiries. The reason for array is because a question can have multiple agencies so the enquiry should hit all of them. In the PR it is set to just pick the first one to reduce number of changes (the same agency shown in the back button).
- Add test for `mailservice`, `enquiryservice`
- Have the enquiry modal use the endpoint.

The usage of `mailservice` was done within `enquiryservice` rather than `enquirycontroller` as per convention in anticipation of future iterations of enquiries not being sent as email but stored in a database so that we can analyse them or add them as questions.

## Before
image

## After
image
image

Deps:
Added `sqlite3` in devDep

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -35,6 +35,7 @@
     "mysql2": "^2.3.0",
     "nodemailer": "^6.4.17",
     "sequelize": "^6.3.5",
+    "sqlite3": "^5.0.2",
     "triple-beam": "^1.3.0",
     "winston": "^3.3.3"
   },
--- a/server/package.json
+++ b/server/package.json
@@ -35,7 +35,6 @@
     "mysql2": "^2.3.0",
     "nodemailer": "^6.4.17",
     "sequelize": "^6.3.5",
-    "sqlite3": "^5.0.2",
     "triple-beam": "^1.3.0",
     "winston": "^3.3.3"
   },
@@ -66,6 +65,7 @@
     "jest": "^27.0.6",
     "lint-staged": "^11.1.2",
     "prettier": "^2.2.1",
+    "sqlite3": "^5.0.2",
     "ts-jest": "^27.0.4",
     "ts-node-dev": "^1.1.6",
     "typescript": "^4.2.3"
--- a/server/src/bootstrap/index.ts
+++ b/server/src/bootstrap/index.ts
@@ -16,6 +16,7 @@ import { PostController } from '../modules/post/post.controller'
 import { AuthController } from '../modules/auth/auth.controller'
 import { TagsController } from '../modules/tags/tags.controller'
 import { MailService } from '../modules/mail/mail.service'
+import { EnquiryController } from '../modules/enquiry/enquiry.controller'
 import { AgencyService } from '../modules/agency/agency.service'
 import { UserService } from '../modules/user/user.service'
 import { AuthService } from '../modules/auth/auth.service'
@@ -33,7 +34,8 @@ import { requestLoggingMiddleware } from './logging'
 
 import { helmetOptions } from './helmet-options'
 import { emailValidator } from './email-validator'
-
+import { EnquiryService } from '../modules/enquiry/enquiry.service'
+import { Agency as agency } from './sequelize'
 export { sequelize } from './sequelize'
 export const app = express()
 
@@ -67,15 +69,17 @@ const mailOptions = {
 }
 const transport = createTransport(mailOptions)
 
+const agencyService = new AgencyService()
 const authService = new AuthService({ emailValidator, jwtSecret })
 const authMiddleware = new AuthMiddleware({ jwtSecret })
 const mailService = new MailService({
   transport,
   mailFromEmail: mailConfig.senderConfig.mailFrom,
 })
+const enquiryService = new EnquiryService({ agency, mailService })
 
 const apiOptions = {
-  agency: new AgencyController({ agencyService: new AgencyService() }),
+  agency: new AgencyController({ agencyService }),
   answers: {
     controller: new AnswersController({
       authService,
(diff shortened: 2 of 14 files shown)


server/src/bootstrap/index.ts
```diff
@@ -108,6 +110,7 @@ const apiOptions = {
     }),
     authMiddleware: authMiddleware,
   },
+  enquiries: new EnquiryController(mailService, agencyService),
```
LoneRifle: Follow convention and supply as...
```suggestion
  enquiries: new EnquiryController({ mailService, agencyService }),
```

server/src/modules/enquiry/enquiry.controller.ts
```diff
@@ -0,0 +1,61 @@
...
+  constructor(
+    private mailService: Public<MailService>,
+    private agencyService: Public<AgencyService>,
+  ) {}
+
+  postEnquiry: ControllerHandler<
+    unknown,
+    unknown,
+    {
+      agencyID: Array<string>
```
LoneRifle:
```suggestion
      agencyId: Array<string>
```
and elsewhere, to be in line with conventions that `id` is stylised as lowercase in this codebase

server/src/modules/enquiry/enquiry.controller.ts
```diff
@@ -0,0 +1,61 @@
...
+      if (!validData) {
+        return res
+          .status(404)
+          .json({ message: 'One or more agency IDs are invalid' })
+      }
+
+      const enquiry = req.body.enquiry
+      enquiry.questionTitle = 'AskGov: ' + enquiry.questionTitle
+
+      await this.mailService.sendEnquiry({ agencyEmail, ccEmail, enquiry })
```
LoneRifle: There is quite a fair bit of biz logic that goes on here. Bear in mind also that this will change in later iterations, so we probably do not want to be messing around with the controller then.

I think it would be prudent if all this is done within an EnquiryService that in turn delegates to:
- Either AgencyService or the Sequelize model for Agency to look up the agency (the latter preferred), and;
- MailService, to send the mail.

I realise that this breaks with conventional wisdom that it's okay for the controller to pass to multiple services, but I'm mindful that how enquiries are sent in may change over time.

LoneRifle (changes requested):

server/src/modules/enquiry/enquiry.controller.ts
nauynix: Noted, I added EnquiryService to use MailService and Agency sequalize model.

LoneRifle (changes requested):
lgtm so far. Could you provide unit tests for the enquiry service ~and controller~ please?

server/src/modules/enquiry/enquiry.service.ts
```diff
@@ -0,0 +1,40 @@
...
+  }): Promise<void> => {
+    const agencyEmail = (await Promise.all(
+      agencyId.map(
+        async (Id) =>
+          (
+            await Agency.findOne({
+              attributes: ['email'],
+              where: { Id },
+            })
+          )?.email ?? '',
```
LoneRifle: This is difficult to read. Could you write it out properly please? Also figure a way to avoid the cast to `Array<string>` (really `string[]` might be more concise)

server/src/modules/enquiry/__tests__/enquiry.service.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+  defineTag,
+  defineUserAndPermission,
+} from '../../../models'
+import minimatch from 'minimatch'
+
+const mockedSequelize = new Sequelize({
+  dialect: 'sqlite',
+  username: 'username',
+  logging: false,
+})
```
LoneRifle: `mockedSequelize` isn't entirely accurate; you are using an actual sequelize instance, just that it's using sqlite

server/src/modules/enquiry/__tests__/enquiry.service.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+    it('should return error when a agency ID is invalid', async () => {
+      // Arrange
+      const agencyId = ['1234']
+      jest.spyOn(agency, 'findOne').mockResolvedValueOnce(null)
+
+      // Act
+      // Assert
+      await expect(
+        enquiryService.emailEnquiry({ agencyId, enquiry }),
+      ).rejects.toThrowError('One or more agency IDs are invalid')
```
LoneRifle: Act and Assert on two separate lines please

server/package.json
```diff
@@ -35,6 +35,7 @@
     "mysql2": "^2.3.0",
     "nodemailer": "^6.4.17",
     "sequelize": "^6.3.5",
+    "sqlite3": "^5.0.2",
```
LoneRifle: Move this to dev dependencies - it's not needed in prod

LoneRifle (changes requested):
lgtm otherwise

LoneRifle (approved):

PR #45, opened 2021-08-10
fix(convict): allow logoHost to be configured

LoneRifle:
## Problem

The logo host is currently hardcoded to S3 on AWS. This is problematic when trying to specify another host, 
say one behind a CDN, for the same

## Solution

Allow `baseConfig.logoHost` to be configured by `LOGOS_HOST` env var


nauynix (approved):
Lgtm

PR #60, opened 2021-08-10
build(deps): schedule dependabot at 1am SGT

LoneRifle:
## Problem and Solution

Reschedule dependabot to 1am, have mergify focus on test-related statuses

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -4,7 +4,6 @@ pull_request_rules:
       - author~=^dependabot(|-preview)\[bot\]$
       - check-success~=Lint
       - check-success~=Test
-      - check-success~=Build and push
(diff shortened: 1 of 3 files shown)


nauynix (approved):
lgtm

.github/mergify.yml
```diff
@@ -4,7 +4,6 @@ pull_request_rules:
       - author~=^dependabot(|-preview)\[bot\]$
       - check-success~=Lint
       - check-success~=Test
-      - check-success~=Build and push
```
mantariksh, line 7: is this strict enough to ensure that both the frontend and backend can build before the upgrade can be merged?

LoneRifle: Added one more step for linting, please review

PR #79, opened 2021-08-11
chore(docker): introduce localstack, maildev via docker-compose

### LINKED ISSUE #21: Use docker compose to set up local db
Downloading and setting up sql locally is time consuming for new developers and has been known to have issues.

Use docker compose instead for a smoother developer experience, and to allow for the addition of future services in future (eg, localstack and maildev)

LoneRifle:
As AskGov grows, so will the sophistication of its architecture and
hence its tooling. To cater to this, introduce docker-compose to spin
up the supporting cast of services that AskGov will depend on.

This initial commit introduces localstack for S3 facilities, and
moves maildev to docker. Future changes may involve introducing more
services, eg, Elasticsearch for text search.

- Create `docker-compose.yml`, spinning up localstack and maildev
  and exposing their ports to the host
- Add `.localstack/` to .gitignore
- Replace `maildev` with `docker-compose up` in `npm run dev`, and
  remove the maildev npm package since we now use it via docker

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 **/node_modules
 /.idea
 /.vscode
+/.localstack
 .env
 .env.*
 !.env.example
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,19 @@
+# This docker-compose file is only for local development purposes.
+version: '3'
+services:
+  localstack:
+    image: localstack/localstack
+    ports:
+      - '4566:4566'
+    environment:
+      - SERVICES=s3
+      - DATA_DIR=/tmp/localstack/data
+    volumes:
+      - './.localstack:/tmp/localstack'
+      - '/var/run/docker.sock:/var/run/docker.sock'
+
+  maildev:
+    image: maildev/maildev
+    ports:
+      - '1080:80'
+      - '1025:25'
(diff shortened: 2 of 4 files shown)


mantariksh (approved):
adding my thoughts on dockerising the application for the record:

actually I do see advantages to running the app in a container:
- makes the dev environment more similar to production
- easier to specify all the config and see the relationships between diff containers in one place (`docker-compose`). actually I see this as the main advantage of dockerising rn - it would be easier to just run localstack on bare metal and just add a `&& localstack start` to our `npm run dev` script, but this isn’t sustainable as our config becomes more complex.

but there are also drawbacks (e.g. need to rebuild the stupid container every time you npm install something new) so I’m ok either way, just putting these thoughts here for the record.

LoneRifle:
@mantariksh - hmm. I think that running localstack on bare metal is a _disadvantage_, since that would imply a global pip install =)

I largely agree with your points of having the config and environment described from one location, but this assumes that the developer comes in already understanding how docker-compose works. Given that this is one step removed from understanding how processes work on bare metal, the upside from docker-compose is largely realised only when a dev team is firmly established and norms are in place. The team behind AskGov is quite transient right now, and given that it is generally easier for a new/temporary engineer to reason about `npm run dev` when close to everything is run on bare metal, I'm more comfortable that docker-compose is limited to spinning up the ecosystem that support the application, but not the application itself.

PR #83, opened 2021-08-12
fix: remove duplicate view count on posts

### LINKED ISSUE #33: questions view count is buggy
helpgovsg/server/modules/post/post.service.ts
Lines 312 to 322 in c23acfb

`views` currently increment on every GET of a post. Issues:

- If the same user navigates multiple times to a page, `views` increments
- We refetch questions from the server to keep the data fresh eg. if someone navigates away from the browser then returns, we GET the post again.

nauynix:
## Problem

Closes #33

## Solution

Pass in `post` as a prop instead of fetching post again in `PostCell`

## Before

https://user-images.githubusercontent.com/20250559/129227742-7a6cc66b-c645-41d2-8840-40b7d759bc19.mov

## After
https://user-images.githubusercontent.com/20250559/129227682-9be22235-066f-4d61-9bfa-6163f94c3286.mov

### DIFF
--- a/client/src/pages/Post/Post.component.jsx
+++ b/client/src/pages/Post/Post.component.jsx
@@ -120,7 +120,7 @@ const Post = () => {
           ) : null}
         </div>
         <div className="question-main">
-          <QuestionSection postId={postId} />
+          <QuestionSection post={post} />
           <AnswerSection post={post} />
         </div>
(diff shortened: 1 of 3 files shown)


LoneRifle (approved):
Lgtm, good catch

PR #84, opened 2021-08-12
fix(authprovider): catch spurious unauthorized error

### LINKED ISSUE #14: Uncaught error in `AuthProvider.whoami()` when unauthenticated
When a user is not authenticated, `axios` throws an error that is uncaught. While there is no noticeable effect on the user, for developers working with Firefox it can be noticeably disruptive. This error should be duly caught and handled.

nauynix:
## Problem

Closes #14

## Solution

Catch any error with status 401, continue to throw everything else

## Before

https://user-images.githubusercontent.com/20250559/129229569-88a88c41-23c7-47e2-b123-3a18657cc929.mov

## After

https://user-images.githubusercontent.com/20250559/129229511-87be392c-24d2-4765-8b47-b50e5902d1a2.mov

### DIFF
--- a/client/src/contexts/AuthContext.tsx
+++ b/client/src/contexts/AuthContext.tsx
@@ -1,3 +1,4 @@
+import { AxiosError } from 'axios'
 import React, { createContext, useContext, useEffect, useState } from 'react'
 import { useMutation, UseMutationResult } from 'react-query'
 import { ApiClient } from '../api'
@@ -37,6 +38,12 @@ export const AuthProvider = ({
       .then((data) => {
         if (data) {
           setUser(data)
+        }
+      })
+      .catch((reason: AxiosError) => {
+        // Catch 401 which signals an unauthorized user, which is not an issue
+        if (!(reason.response && reason.response.status === 401)) {
+          throw reason
         }
       })
   }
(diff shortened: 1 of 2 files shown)


LoneRifle (changes requested):
Hmm, there may be 400 errors too arising from malformed tokens (use Google Chrome to verify this), so you'll need to take those into account too.

Additionally, I don't think you want to rethrow the other errors since the user then cannot act on it. At the very least, log it, or maybe indicate to user with a toast.

nauynix:
> Additionally, I don't think you want to rethrow the other errors since the user then cannot act on it. At the very least, log it, or maybe indicate to user with a toast.

I only wanted to catch the 401 errors so the behaviour of the other errors remain the same. I think if we wanted to catch and present the other errors, it should be in another PR?

nauynix:
I think this is being approached the wrong way. `401` is an error, which is why `axios` is throwing it. But its not really an error; its just saying that the user is unauthorized (which the majority of users are), so it should return something else.

LoneRifle:
The issue really relates to errors being uncaught. Given that the changes are small enough, you might as well address both the expected unauthz errors as well as the unexpected ones in the same PR.

May I also point out that Chrome users will experience 400 errors instead of 401 owing to how the token is blanked out in localStorage instead of properly deleted.

nauynix:
May I suggest changing all the unauthorised response codes in from the server to 401 since they are all of the same nature then?

LoneRifle (approved):
That works for now - please follow up with the toast as suggested

PR #88, opened 2021-08-13
Fix/ampersand in query

### LINKED ISSUE #10: Clicking on Career & Scholarships tag results in invalid tag error
First take suggests it is an encoding issue as other multi word tags like `TOPIC TAG 1` work correctly when clicking on tag to see all questions under topic tag as discovered by @barnettx 

https://user-images.githubusercontent.com/18087283/119076280-ddefc780-ba24-11eb-9024-7a4d4deeb60e.mov

image

Replacing `&` with `%26` in the query string fixes it.

nauynix:
## Problem

Closes #10

## Solution

Encode all agency and tag names in url. Agencies as well in case of a future agency with '&' in name.
## Before
image

## After
image

### DIFF
--- a/client/src/pages/Post/Post.component.jsx
+++ b/client/src/pages/Post/Post.component.jsx
@@ -146,7 +146,11 @@ const BackButtonForPostDetail = ({ agencyShortName }) => {
 
   return (
     <div className="back-to-home">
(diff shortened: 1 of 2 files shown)


LoneRifle (approved):
lgtm

PR #94, opened 2021-08-16
refactor(enquiries): use Sequelize convention for Agency model

### LINKED ISSUE #17: User can fill a form to ask questions to agency

LoneRifle:
## Problem and Solution

Per PR title, adhere to conventions established by Sequelize with respect to how models are used


nauynix (approved):
Lgtm

PR #99, opened 2021-08-17
refactor(user): remove `responseHandler`, make login() sync

### LINKED ISSUE #5: Remove `responseHandler` from UserService
`responseHandler` provides very little value, and the patterns involving its use in the codebase are dated and confusing. Remove responseHandler and its invocations, reshaping responses as appropriate.

LoneRifle:
## Problem

Closes #5 

## Solution

To facilitate testability and cleanly separate controller/biz concerns,
remove `responseHandler` from UserService, and take opportunities to
clean up where convenient

[UserService]
- remove `responseHandler` from `loadUser()`
- remove unnecessary awaits
- simplify `login()` to be synchronous

[AuthController]
- type `loadUser()` as ControllerHandler
- remove use of `responseHandler` across all functions

[AuthContext]
- change expected response shape when calling `/auth`

### DIFF
--- a/server/src/modules/auth/auth.controller.ts
+++ b/server/src/modules/auth/auth.controller.ts
@@ -1,7 +1,7 @@
-import helperFunction from '../../helpers/helperFunction'
 import { validationResult } from 'express-validator'
 import { generateRandomDigits, hashData, verifyHash } from '../../util/hash'
 import { User, Token } from '../../bootstrap/sequelize'
+import { User as UserType } from '../../models'
 import { createValidationErrMessage } from '../../util/validation-error'
 
 import { MailService } from '../mail/mail.service'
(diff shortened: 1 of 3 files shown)


server/src/modules/auth/auth.controller.ts
```diff
@@ -33,27 +35,18 @@ export class AuthController {
     this.userService = userService
   }
 
-  loadUser = async (req: Request, res: Response): Promise<Response> => {
+  loadUser: ControllerHandler<unknown, UserType | null | Message> = async (
```
nauynix, line 38: Why is there the need for null?

LoneRifle: Sequelize and hence `UserService.loadUser()` could possibly return null

nauynix (approved):
lgtm

PR #101, opened 2021-08-17
refactor(message-type): rename, use in enquiries

LoneRifle:
## Context
Ensure responses are typed by specifying the Message type in the appropriate type parameter for ControllerHandler


nauynix (commented):
Rename error-message-type? In this case its not returning an error

nauynix (approved):
lgtm

PR #102, opened 2021-08-17
feat(enquiry): add recaptcha

### LINKED ISSUE #17: User can fill a form to ask questions to agency

nauynix:
Adds recaptcha to enquiry form. Server recaptcha code under `services` is adapted from `formSG`.
The two new env variables are test keys given by google in the [public](https://developers.google.com/recaptcha/docs/faq#id-like-to-hide-the-recaptcha-badge.-what-is-allowed) 

Closes #17 

## Before & After Screenshots

**BEFORE**:
image

**AFTER**:
image

- `REACT_APP_RECAPTCHA_SITE_KEY` : Recaptcha key for client to generate user response
- `RECAPTCHA_SECRET_KEY` : Recaptcha key for server to verify user response

**New dependencies**:

`Client`
- `react-google-recaptcha`

`Server`
- `axios` : to verify user response for recaptcha
- `https-status-codes` : since the code for `formSG` uses this package too, good to add it now and change all the other handwritten status code 
- `neverthrow` : since the code for `formSG` uses this package too, good to add it now and change all the other return values in `services` to be `result`\`resultAsync`
- 'supertest` : to test `enquiry.routes` and `enquiry.controller`.


server/src/services/recaptcha/recaptcha.service.ts
```diff
@@ -0,0 +1,55 @@
...
+
+import { GOOGLE_RECAPTCHA_URL } from './recaptcha.constants'
+import {
+  CaptchaConnectionError,
+  MissingCaptchaError,
+  VerifyCaptchaError,
+} from './recaptcha.errors'
+
+const logger = createLogger(module)
+export const verifyCaptchaResponse = (
```
LoneRifle: For consistency, we should have a RecaptchaService which accepts an axios client in its ctor. This would allow you to avoid the module mocking in your tests. Since you would also be injecting the RecaptchaService into the EnquiryController, you can also provide unit-testing for the controller and even the route (especially given that you now have supertest).

For convenience, the ctor can also accept just the Google ReCAPTCHA URL; you then construct an axios client from that

server/src/services/recaptcha/recaptcha.constants.ts
```diff
@@ -0,0 +1,2 @@
+export const GOOGLE_RECAPTCHA_URL =
+  'https://www.google.com/recaptcha/api/siteverify'
```
LoneRifle: should be provided as a convict config whose sole value (or default) is this

server/src/services/recaptcha/__tests__/recaptcha.service.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+
+import { GOOGLE_RECAPTCHA_URL } from '../recaptcha.constants'
+import {
+  CaptchaConnectionError,
+  MissingCaptchaError,
+  VerifyCaptchaError,
+} from '../recaptcha.errors'
+import { verifyCaptchaResponse } from '../recaptcha.service'
+
+const MOCK_PRIVATE_KEY = recaptchaConfig.recaptchaSecretKey
```
LoneRifle: you're using mocks, so there's no need to reference bootstrap here.

LoneRifle (changes requested):
The biz logic is sound, having tested on dev and staging. Some feedback given concerning refactoring and providng coverage tests. Good work nonetheless!

server/src/bootstrap/config/recaptcha.ts
```diff
@@ -0,0 +1,26 @@
...
+  recaptchaSecretKey: string
+}
+
+const recaptchaSchema: Schema<RecaptchaConfig> = {
+  recaptchaSiteKey: {
+    doc: 'Recaptcha key for client to generate user response',
+    format: String,
+    default: '',
+    env: 'REACT_APP_RECAPTCHA_SITE_KEY',
+  },
```
LoneRifle: this is not actually needed here, since it's webpacked into the frontend

server/src/services/recaptcha/recaptcha.service.ts
nauynix: Going with the first suggestion of passing an axios into RecaptchaService

server/src/modules/enquiry/__tests__/enquiry.routes.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+  defineTag,
+  defineUserAndPermission,
+} from '../../../models'
+import { RecaptchaService } from '../../../services/recaptcha/recaptcha.service'
+import { MailService } from '../../mail/mail.service'
+import { EnquiryController } from '../enquiry.controller'
+import { routeEnquiries } from '../enquiry.routes'
+import { EnquiryService } from '../enquiry.service'
+
+describe('/enquires', () => {
```
LoneRifle:
```suggestion
describe('/enquiries', () => {
```

LoneRifle (approved):
lgtm otherwise, merge when ready

PR #103, opened 2021-08-17
Fix/remove response handler

### LINKED ISSUE #106: Remove `responseHandler` from all modules
`responseHandler` provides very little value, and the patterns involving its use in the codebase are dated and confusing. Remove responseHandler and its invocations, reshaping responses as appropriate.

nauynix:
## Problem

## Solution

This PR is purely a refactor and only removes `responseHandler`, and does not shap the responses of controllers to `ControllerHandler`. Since some removal of `responseHandler` also affect the frontend (`data.data` to `data`), it is beneficial to review PR by commits.


server/src/modules/answers/answers.controller.ts
```diff
@@ -68,35 +59,16 @@ export class AnswersController {
...
-            function: 'addAnswer',
-            userId: req.user.id,
-            postId: req.params.id,
-          },
-          error,
-        })
-        return res.status(error.code).json(error)
-      }
-      return res.status(data?.code || 200).json(data)
+      return res.status(201).json(data)
```
LoneRifle, line 68: since the scope of the PR is very wide, avoid taking the opportunity to change biz logic

server/src/modules/answers/answers.controller.ts
```diff
@@ -159,29 +104,14 @@ export class AnswersController {
...
-              answerId: req.params.id,
-            },
-            error,
-          })
-          return res.status(error.code).json(error)
-        }
-        // Assuming that if err returned null then data is defined
-        return res.status(data!.code).json(data)
-      })
+      this.answersService.remove(req.params.id)
```
LoneRifle, line 113:
```suggestion
      await this.answersService.remove(req.params.id)
```

server/src/modules/post/post.controller.ts
```diff
@@ -31,25 +31,24 @@ export class PostController {
...
-        meta: {
-          function: 'listPosts',
-        },
-        error,
-      })
-      return res
-        .status(500)
-        .json(helperFunction.responseHandler(true, 500, 'Server Error', null))
+      if (error === 'Invalid tags used in request') {
+        return res.status(422).json({ message: error })
```
LoneRifle, line 40: Again, avoid conflating changes relating to returning the correct HTTP status codes, given that the focus of this PR is removing `responseHandler`

server/src/modules/post/post.controller.ts
```diff
@@ -75,18 +64,16 @@ export class PostController {
...
 
     try {
       await this.authService.verifyUserCanViewPost(
-        post.toJSON(),
+        post,
         req.header('x-auth-token') ?? '',
       )
     } catch (error) {
-      logger.error({
+      logger.warn({
```
LoneRifle, line 75: Why did you change this?

server/src/modules/post/post.service.ts
```diff
@@ -343,20 +344,12 @@ export class PostService {
           'answer_count',
         ],
       ],
-    })
+    })) as PostWithUserRelations
```
LoneRifle, line 290: There is little need to do a cast here, since `post` can be considered a subtype of PostWithUserRelations.

See Liskov's Principle of Substitution.

server/src/modules/post/post.service.ts
```diff
@@ -212,31 +212,15 @@ export class PostService {
...
-      return [
-        helperFunction.responseHandler(
-          false,
-          400,
-          'At least one tag does not exist',
-          null,
-        ),
-        null,
-      ]
+      return Promise.reject('At least one tag does not exist')
```
LoneRifle: As discussed, we should favour throwing Errors instead of using `Promise.reject()`

server/src/modules/post/post.service.ts
```diff
@@ -240,21 +240,15 @@ export class PostService {
...
     const update = await PostModel.update(
       { status: 'ARCHIVED' },
       { where: { id: id } },
     )
     if (!update) {
-      return [
-        helperFunction.responseHandler(false, 400, 'Update failed', null),
-        null,
-      ]
+      Promise.reject('Update failed')
```
LoneRifle: This is not being returned, so it's effectively a no-op

LoneRifle (changes requested):
I did not review all commits. There is a good reason why wide-ranging PRs like these are rare and usually treated very carefully. Such undertakings are not to be done rashly!

This one is doing too much in one go, conflates concerns about changing HTTP status codes, and - unlike what is offered in the PR's description - does not actually make corresponding changes in the frontend, except for two particular instances.

Redo these changes as a series of PRs, taking into account the feedback above.

PR #109, opened 2021-08-19
fix(enquiry): actually await on `EnquiryService.emailEnquiry()`

### LINKED ISSUE #17: User can fill a form to ask questions to agency

LoneRifle:
## Problem

A bug was uncovered during testing, where the backend rejected the sending of emails, 
but was also reporting successful sending of enquiry to the user

## Solution

Given that the abovementioned function is asynchronous, ensure that
we await on it, updating the test suite to reflect this

- Ensure that mocks of `EnquiryService.emailEnquiry()` either resolves
  or rejects
- Ensure that we await on the async `EnquiryService.emailEnquiry()`


nauynix (approved):
lgtm

PR #112, opened 2021-08-19
fix(enquiry): update description to 3-14 working days

nauynix:
## Solution
Updated description to say 3-14 working days instead of at least 3 days.
## BEFORE
image
## AFTER
image


LoneRifle (approved):
lgtm

PR #115, opened 2021-08-19
Chore(post): remove responseHandler

nauynix:
## Problem

## Solution
To keep changes small, this PR only removes responseHandler under `posts`

### DIFF
--- a/server/src/modules/post/post.controller.ts
+++ b/server/src/modules/post/post.controller.ts
@@ -31,25 +31,24 @@ export class PostController {
     const { query } = req
     const { sort = SortType.Top, tags = '' } = query
     try {
-      const [err, data] = await this.postService.retrieveAll({
+      const data = await this.postService.retrieveAll({
         sort: sort as SortType,
         tags: tags as string,
       })
-      if (err) {
-        return res.status(err.code).json(err)
-      }
-      return res.status(data?.code || 200).json(data?.data)
+      return res.status(200).json(data)
     } catch (error) {
-      logger.error({
-        message: 'Error while listing posts',
-        meta: {
-          function: 'listPosts',
-        },
-        error,
-      })
-      return res
-        .status(500)
-        .json(helperFunction.responseHandler(true, 500, 'Server Error', null))
+      if (error === 'Invalid tags used in request') {
+        return res.status(422).json({ message: error })
+      } else {
+        logger.error({
+          message: 'Error while listing posts',
+          meta: {
+            function: 'listPosts',
+          },
+          error,
+        })
+        return res.status(500).json({ message: 'Server Error' })
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


server/src/modules/post/post.service.ts
```diff
@@ -343,20 +316,12 @@ export class PostService {
           'answer_count',
         ],
       ],
-    })
+    })) as PostWithUserRelations
```
nauynix, line 291: Reason for typing as PostWithUserRelation is because `authService.verifyUserCanViewPost` accepts only PostWithUserRelation in the controller function that calls this func.

server/src/modules/post/post.service.ts
```diff
@@ -381,15 +318,7 @@ export class PostService {
...
-      return [
-        helperFunction.responseHandler(
-          false,
-          422,
-          'Invalid tags used in request',
-          null,
-        ),
-        null,
-      ]
+      throw 'Invalid tags used in request'
```
LoneRifle: Never throw a literal, always throw Errors instead. I'm trying to figure out why our linting rules don't include this.
See https://eslint.org/docs/rules/no-throw-literal for more info.

server/src/modules/post/post.service.ts
```diff
@@ -141,28 +125,17 @@ export class PostService {
...
   filterPostsWithoutAnswers = async (
     data?: PostWithRelations[],
-  ): Promise<HelperResult> => {
+  ): Promise<Post[]> => {
     if (!data) {
-      return [
-        helperFunction.responseHandler(false, 404, 'No posts found', null),
-        null,
-      ]
+      return Array<Post>()
```
```suggestion
      return []
```

LoneRifle (changes requested):
lgtm otherwise

server/src/modules/post/post.controller.ts
```diff
@@ -31,43 +31,31 @@ export class PostController {
...
-        .json(helperFunction.responseHandler(true, 500, 'Server Error', null))
-    }
-  }
-
-  getSinglePost = async (req: Request, res: Response): Promise<Response> => {
-    let post
-    try {
-      const [error, data] = await this.postService.retrieveOne(req.params.id)
-      if (error) {
+      if (error === 'Invalid tags used in request') {
```
nauynix: Literal string was thrown to match this. If throwing an Error obj change this too.

server/src/modules/post/post.controller.ts
```diff
@@ -37,7 +37,7 @@ export class PostController {
       })
       return res.status(200).json(data)
     } catch (error) {
-      if (error === 'Invalid tags used in request') {
+      if (error.message === Error('Invalid tags used in request').message) {
```
LoneRifle: `error.message` is a String, so you don't have to wrap a string literal in an Error only to access it via `message`

LoneRifle (approved):

PR #117, opened 2021-08-19
Chore(auth): remove responseHandler

nauynix:
## Problem

## Solution
To keep changes small, this PR only removes responseHandler under `auth`


LoneRifle (approved):
lgtm

PR #127, opened 2021-08-23
refactor(helpers): remove all traces of helperFunction

LoneRifle:
## Problem and Solution

Remove helperFunction now that it is no longer used


nauynix (approved):
lgtm

PR #129, opened 2021-08-23
Feat/hamburger

nauynix:
## Problem

Mobile and tablet view user do not have a way to view and select tags, as the tag panel only shows up above `768px`.

## Solution

Add a hamburger menu using the new Chakra `menu` component, following the AskGov Design layout.
Only show for devices with width smaller than `768px`.

## Before & After Screenshots

**BEFORE**:

https://user-images.githubusercontent.com/20250559/130393397-9920ee93-75c2-42ca-a2cf-d44ad59d4e55.mov

**AFTER**:
Mobile

Tablet
Screen Shot 2021-08-24 at 11 23 28
Desktop
Screen Shot 2021-08-24 at 11 23 40


client/src/components/SearchBox/SearchBox.component.jsx
```diff
@@ -97,6 +97,7 @@ const SearchBox = ({ placeholder, value, handleSubmit }) => {
             <div className="search-autocomplete">
               <InputGroup className="search-box">
                 <InputLeftElement
+                  zIndex="0"
```
nauynix, line 100: By default, the zIndex of `InputLeftElement` is 2 and will clip the menu dropdown, even if portalled. Set to 0 so the search icon wont clip through.

LoneRifle:
First glance suggests that the hamburger menu is too far away from the content of relevance. @khaleedah wdyt?

nauynix:
@khaleedah agreed that there will be major changes to the design of the menu and header. My take is that we go through with this first to provide mobile users *some* way of choosing tags, and replace it when the design team has a finalised version.

LoneRifle:
As discussed, we'll not proceed with this for now. The topic selector should be situated closer to the content it's filtering, and a hamburger situated in the top-left may mislead users into thinking that it contains app-wide changes.

Let's go with @khaleedah's proposed changes to the topic dropdown selectors instead.

PR #141, opened 2021-08-25
Feat: home page lists posts in pages

### LINKED ISSUE #20: Q&A is not paginated

nauynix:
## Problem
Homepage posts are not paginated, leading to two issues. 
1. the size of response is quite big as server is returning whole list of post, and will only grow with number of posts
2. The page becomes very long and tedious to scroll through

- Closes #20

## Solution

This PR adds the ability to query for `posts` and `answerable posts` with two new request querys: `page` and `size`, which are optional. If they are not specified, the server returns the whole list. 

- `/posts` and `/posts/answerable` used to return `posts`, now returns shape `{posts, totalItems}`
- Take the opportunity to correct mistakes in documentation 
- Remove superseded and unused routes in `post.routes`. The functionality of `/top` and `/tag/:tagname` is covered in `/`
- Update `post` endpoint in `README`
- Update frontend to use new query, cache previous results to reduce requests when navigating back to previous page

## Screenshots
image


server/src/modules/post/post.service.ts
```diff
@@ -41,17 +41,44 @@ export class PostService {
     return ['views', 'DESC']
   }
 
+  private getPaginatedPosts = (
+    post: Post[],
```
LoneRifle: nitpick
```suggestion
    posts: Post[],
```
and elsewhere

server/src/modules/post/post.service.ts
```diff
@@ -117,12 +144,16 @@ export class PostService {
...
-      return []
+      return { posts: [], totalItems: 0 }
     } else if (withAnswers) {
-      return returnPosts
+      return this.getPaginatedPosts(returnPosts, page, size)
     } else {
       // posts without answers
-      return this.filterPostsWithoutAnswers(posts)
+      return this.getPaginatedPosts(
+        await this.filterPostsWithoutAnswers(posts),
```
LoneRifle: nit - just as good practice and to make it clearer that we await on this before moving to `this.getPaginatedPosts()`, I would have this on its own line.

LoneRifle (changes requested):
Couple more points:
- Provide _some_ tests for at least post.service, at least for the functions you changed
- Current page size is hardcoded to 10, which I've personally found to be quite limiting. Raising that limit to 50 may provide a better balance of number of pages to click through vs resulting response size.

lgtm otherwise

LoneRifle (approved):
nicely done

PR #148, opened 2021-08-27
feat: add seq script to seed data from csv

nauynix:
## Problem and Solution

When importing the questions to IPOS previously, a sequalize script was written to run it, but it was never added to the codebase. Add it now to aid in future importing of other agencies.


LoneRifle (approved):
lgtm

PR #162, opened 2021-08-30
Chore(test): insert data into test database instead of mocking it

nauynix:
## Problem

All tests that makes calls to query the database were previously implemented by mocking the database return value. This means that any code that makes queries will not be tested and hence, the test has room to be more comprehensive.

## Solution
- New function that creates a test database to insert values
- Change all tests that previously mocked data to now create the data in the
test database


mantariksh (commented):
makes sense for integration tests, but for the service unit tests imo it would make more sense to mock the database layer. that's the idea of a unit test - it tests one small, specific piece of functionality (in this case, "does the service call the database layer correctly?") so that if the test fails, the breaking change is immediately obvious. @LoneRifle thoughts?

server/src/bootstrap/config/auth.ts
```diff
@@ -9,7 +9,7 @@ const authSchema: Schema<AuthConfig> = {
   jwtSecret: {
     doc: 'JWT authentication secret',
     format: String,
-    default: null,
+    default: 'toomanysecrets',
```
LoneRifle: there shouldn't be a default here... in production, if there is no value, we deliberately want convict to fail.

nauynix:
My concern is that if we mock the database return, then we are unable to test if it calls the database with the correct query string within the `service` function. Someone could accidentally modify it and the test would not catch it since its going to override it anyway.
https://github.com/opengovsg/askgovsg/blob/3a5b42a4364acb12bc10c50b9786f9eb6c1135cd/server/src/modules/post/post.service.ts#L194-L222

server/src/util/jest-db.ts
```diff
@@ -0,0 +1,52 @@
...
+export enum ModelName {
+  Answer = 'answer',
+  Agency = 'agency',
+  Post = 'post',
+  PostTag = 'posttag',
+  Tag = 'tag',
+  User = 'user',
+  Permission = 'permission',
+  Token = 'token',
+}
```
LoneRifle, line 21: This can (and should) be used by `models/` too, but perhaps in a future commit.

LoneRifle (changes requested):
lgtm otherwise

server/src/bootstrap/config/auth.ts
nauynix: understand, then can we supply github actions with one? Previous commit failed test and lint https://github.com/opengovsg/askgovsg/pull/162/commits/b78f6a16af99442078e3be3d05bf4858be69be44

LoneRifle:
> makes sense for integration tests, but for the service unit tests imo it would make more sense to mock the database layer. that's the idea of a unit test - it tests one small, specific piece of functionality (in this case, "does the service call the database layer correctly?") so that if the test fails, the breaking change is immediately obvious. @LoneRifle thoughts?

I'm a bit more laid back - there is a balance to be struck between testing with realistically behaving dependencies and testing with dependencies whose behaviour you can control. I would generally be happier using the former if I can still control behaviour in some fashion, especially if my understanding of said behaviour in the latter may be prone to error.

server/src/util/jest-db.ts
```diff
@@ -0,0 +1,52 @@
+import { Model, ModelCtor, Sequelize } from 'sequelize'
+import { emailValidator } from '../bootstrap/email-validator'
```
LoneRifle: yea don't bring this in, since it has a dependency on `authConfig`

server/src/bootstrap/config/auth.ts
LoneRifle: have left a comment on the relevant spot that is causing the problem

server/src/util/jest-db.ts
nauynix: replaced emailvalidator with one made for test

LoneRifle (approved):

PR #192, opened 2021-09-01
fix(search): have agency-specific results from agency page

### LINKED ISSUE #178: When at agency page, results from searching and pressing Enter is not agency specific
### To reproduce

- Visit the page for [ParkingSG](https://ask.gov.sg/agency/parkingsg)
- Search for ticket and press Enter

#### Observed
- Search results come from across agencies

#### Expected
- Search results should be specific to agency

LoneRifle:
## Problem

Closes #178 

## Solution

If a user is searching from an agency page, the user expects to see
results specific to that agency, so pass this to the SearchResults
component so that the appropriate posts are fetched for searching

## Screenshot

https://user-images.githubusercontent.com/10572368/131606034-a0dd6384-58c9-4aa0-bd8d-6345d3a46bcb.mov


nauynix (approved):
lgtm

PR #199, opened 2021-09-01
feat(enquiries): suggest existing questions to the user

### LINKED ISSUE #173: when submitting an enquiry form, auto complete as user is typing `Question Title`
this caters to users who did not attempt to search the site before submitting their question.

downside: for a user who has already searched, this is potentially annoying because we are showing the same answers that are not helpful for them.

Given that our primary goal is to reduce repeat questions to agency users, we will build this in and monitor results.

LoneRifle:
## Problem

Closes #173 

## Solution

- refine default handleSubmit flag to just push history; regardless of
  specified hook, GA event will be sent
- add handleAbandon hook, capturing when a search is abandoned
- allow search icon to be disabled by setting `showSearchIcon` to false
- retain search text value even on abandoning
- make SearchBox a component, not a form
- introduce also the means to open search items in new window to retain user context as much as possible
- use SearchBox to suggest questions

## Screenshot

image

### DIFF
--- a/client/src/components/SearchBox/SearchBox.component.jsx
+++ b/client/src/components/SearchBox/SearchBox.component.jsx
@@ -107,11 +107,21 @@ const SearchBox = ({
     keys: ['title', 'description'],
   })
 
+  const stateReducer = (_state, changes) => {
+    return [
+      Downshift.stateChangeTypes.blurInput,
+      Downshift.stateChangeTypes.mouseUp,
+    ].includes(changes.type)
+      ? { isOpen: false } // no-changes
+      : changes
+  }
+
   return (
     <div className="search-container">
       <form className="search-form" onSubmit={handleSubmitHook(onSubmit)}>
         <Downshift
           onChange={(selection) => history.push(`/questions/${selection.id}`)}
+          stateReducer={stateReducer}
           itemToString={itemToString}
         >
           {({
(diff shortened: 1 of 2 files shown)


nauynix:
pressing enter on question title in enquiry modal now leads to search page

LoneRifle:
> pressing enter on question title in enquiry modal now leads to search page

good catch! added a flag to control this behaviour.

nauynix:
image
You can't submit now because it thinks `questionTitle` is empty

LoneRifle:
spotted and corrected the missing `inputRef`

nauynix (approved):
lgtm

client/src/components/SearchBox/SearchBox.component.jsx
```diff
@@ -109,32 +121,38 @@ const SearchBox = ({ placeholder, value, handleSubmit }) => {
...
                     onKeyDown: (event) => {
-                      const selectingOptionUsingKeyboard =
-                        event.key === 'Enter' && highlightedIndex !== null
-                      if (selectingOptionUsingKeyboard) {
+                      if (event.key === 'Enter') {
                         // when selecting option using keyboard
                         // downshift prevents form submission which is used to submit analytics event
                         // detect such event and separately send analytics event
-                        sendSearchEventToAnalytics(inputValue)
+                        if (highlightedIndex === null && searchOnEnter) {
```
nauynix, line 150: nitpick: add comments explaining what the two cases cover. Its also a bit hard to understand because of the `===` and `!==`

nauynix:
```suggestion
                        // hi
                        if (highlightedIndex === null && searchOnEnter) {
```

nauynix:
```suggestion
                        // If user types enter without selecting anything
                        if (highlightedIndex === null && searchOnEnter) {
```

PR #204, opened 2021-09-02
fix(footer): stick enquiry button to bottom of post page

nauynix:
## Problem and Solution
Enquiry button component sticks to the footer of homepage (expected behaviour) but not on postpage.

## Before
Screen Shot 2021-09-02 at 14 35 49

## After
Screen Shot 2021-09-02 at 14 34 51


LoneRifle (approved):
lgtm, good catch

Issue #100, opened 2021-08-17
AuthController: move Sequelize models to respective services


LoneRifle:
To ensure proper separation of concerns between controller and biz logic, AuthService and UserService should hold their relevant Sequelize models. 

The opportunity should also be taken to clean up as follows:
- Controller handlers should be typed
- Unit tests should also be provided for coverage

PR #207, opened 2021-09-03
refactor(user): move `UserService.login` to `AuthService.createToken`

LoneRifle:
## Problem

The UserService should only concern itself with user-related admin
tasks. In addition, AuthService should be the authoritative service 
for handling authentication

Groundwork for #100 

## Solution
Hence, move `login()`, which creates signed JWT tokens, to
AuthService, where it should be. This allows us to simplify our
dependencies by having everything JWT-related in AuthService.


nauynix (approved):
lgtm

PR #216, opened 2021-09-06
build: create directory for shared client and server code

mantariksh:
## Problem

API types should be shared between the frontend and backend to prevent duplication of code. In other words, the frontend service which calls an API and the backend controller which returns the response should share the same type for the shape of data returned.

However, the build pipelines for frontend and backend are currently completely separate. This means code cannot easily be shared between the frontend and backend.

Part of #218 

## Solution

The idea is to rewire the build pipeline to accommodate a `shared` folder at the top level, from which the frontend and backend can both import code.

The changes for the frontend were more involved due to the intricacies of `create-react-app`. The solution was to use the ["Outside of root" section](https://www.npmjs.com/package/react-app-rewire-alias#outside-of-root) of `react-app-rewire-alias`. This allowed us to alias the top-level `shared` folder to `~shared` when importing in the frontend.

The backend was more straightforward and just involved changes to `server/tsconfig.json`.

In order to test that the build pipeline works, I swapped out the existing `BaseUserDto` type which previously lived in `client` and replaced it with a `UserBaseDto` type in `shared`. In the backend, I updated the `User` model type to inherit from `UserBaseDto`. This pattern will be applied to the rest of the models and API types in a subsequent PR.


server/src/models/users.model.ts
```diff
@@ -2,15 +2,11 @@ import { Sequelize, DataTypes, Model, ModelCtor } from 'sequelize'
...
 
 const USER_MODEL_NAME = 'user'
 
-export interface User extends Model {
-  id: string
-  username: string
-  displayname: string
-  views: number
-}
+export interface User extends Model, UserBaseDto {}
```
LoneRifle, line 10: could you leave a comment here to the effect of deprecating the use of Model and ModelCtor in favour of ModelDefined? The idea is that in time, we don't extend Model, instead wrapping around ModelDefined. This would also allow us to use the name User to mean the plain old object, rather than suffixing with Dto.

LoneRifle (approved):
lgtm otherwise

server/src/models/users.model.ts
mantariksh: addressed in https://github.com/opengovsg/askgovsg/pull/216/commits/5f0c96e1efd5dfeb9a0f6d21a84b27d7d9688995

mantariksh:
oops

Screenshot 2021-09-06 at 3 44 05 PM

mantariksh:
had to update path of server's `index.js` for `npm start`, fixed in https://github.com/opengovsg/askgovsg/pull/216/commits/8828f4666f34321a3eac73b592144e721366c6d2 and https://github.com/opengovsg/askgovsg/pull/216/commits/5d4b8dc581a4e87b43e12e4cf1d966ed6a05bfb7

mantariksh:
apologies for force-pushing after review but have to clean up commit history

mantariksh:
@LoneRifle requesting another review for 563d944!

server/tsconfig.json
```diff
@@ -15,5 +14,8 @@
   },
   "include": [
     "src"
+  ],
+  "references": [
+    { "path": "../shared"}
```
LoneRifle:
```suggestion
    { "path": "../shared" }
```

shared/types/user.ts
```diff
@@ -0,0 +1,10 @@
+import { z } from 'zod'
+
+export const UserBaseDto = z.object({
+  id: z.string(),
+  username: z.string(),
+  displayname: z.string(),
+  views: z.number().nonnegative(),
+})
+
+export type UserBaseDto = z.infer<typeof UserBaseDto>
```
LoneRifle: Could we perhaps go with my earlier proposal of simply calling this a User, then alias in server as UserBaseDto as needed?

LoneRifle (approved):
lgtm otherwise

server/tsconfig.json
mantariksh: https://github.com/opengovsg/askgovsg/pull/216/commits/38c7ea0e46abb325608581dd9aed7e705c8cce66

shared/types/user.ts
mantariksh: https://github.com/opengovsg/askgovsg/pull/216/commits/487d91c2aa6ae2bd43e866ac8edc41c86b080448

LoneRifle:
Merge when ready

PR #217, opened 2021-09-06
fix(search): ensure input is preserved on `touchEnd` event

LoneRifle:
## Problem

The SearchBox currently has to manipulate state manually on certain events corresponding to input blur,
to ensure that input remains preserved. This currently listens for `blurInput` and `mouseInput` events, but
not the `touchEnd` event, which fires for mobile devices

## Solution

Listen and handle the `touchEnd` event


nauynix (approved):
lgtm

Issue #206, opened 2021-09-03
Add GA events to record creating and sending enquiries


LoneRifle:
We currently do not know when a user decides to send enquiries to an agency, which is important since that helps us to understand when users are unable to find answers for the questions they have. Add Google Analytics events and trigger them at appropriate points in the user flow in the frontend

PR #219, opened 2021-09-06
fix(search): handle abandon only if searchbox on homepage

nauynix:
## Problem && Solution

`handleAbandon` is fired regardless of `searchbox` in homepage or enquiry form. If it is in enquiry form and user clicks out of search box in order to continue typing, it should not be considered an abandon event.

### DIFF
--- a/client/src/components/SearchBox/SearchBox.component.jsx
+++ b/client/src/components/SearchBox/SearchBox.component.jsx
@@ -121,7 +121,7 @@ const SearchBox = ({
           }) => (
             <div
               className="search-autocomplete"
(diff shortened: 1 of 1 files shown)


nauynix:
As discussed, this change is not needed because we are able to tell which component (homepage vs enquiry form) triggered the event. The ideal user should then trigger the abandon event twice: once when he searches the homepage and fails to find anything, and secondly when he 'searches' the enquiry box. By sending both events we have a more comprehensive overview of user actions.

LoneRifle:
To add even more colour, we currently don't log when the user opens the enquiry form; that's currently an open issue - #206

PR #223, opened 2021-09-06
feat: show related posts

### LINKED ISSUE #19: On Question details page, user can see "Related Questions"
User should be able to view and go to related questions in case the current one does not answer their question, or if they just want to read up more about the topic.

### Logic
show top 3 questions (by popularity) with as many similar tags as possible

nauynix:
## Problem

User should be able to view and go to related questions in case the current one does not answer their question, or if they just want to read up more about the topic.

- Part of #19

## Solution

Find the posts with the most number of shared tags, and order them in terms of number of views. More complex logic to find similar posts is suggested in comments.

## Before & After Screenshots

**BEFORE**:
Screen Shot 2021-09-06 at 14 02 44

**AFTER**:
image

https://user-images.githubusercontent.com/20250559/132167937-8ac9d81b-7835-463a-a98a-3c76fe66dd20.mov

https://user-images.githubusercontent.com/20250559/132167955-1c2daaf3-04dc-43af-8e50-794723767985.mov

### DIFF
--- a/server/src/modules/post/post.controller.ts
+++ b/server/src/modules/post/post.controller.ts
@@ -169,11 +169,15 @@ export class PostController {
    * @return 500 for database error
    */
   getSinglePost: ControllerHandler<
-    { id: string },
+    { id: number },
     PostWithUserTagRelations | PostWithUserTagRelatedPostRelations | Message,
     undefined,
-    { relatedPosts?: string }
+    { relatedPosts?: number }
   > = async (req, res) => {
+    const errors = validationResult(req)
+    if (!errors.isEmpty()) {
+      return res.status(StatusCodes.BAD_REQUEST).json(errors.array()[0].msg)
+    }
     let post
     try {
       post = await this.postService.getSinglePost(
--- a/server/src/modules/post/post.controller.ts
+++ b/server/src/modules/post/post.controller.ts
@@ -172,13 +172,13 @@ export class PostController {
     { id: string },
     PostWithUserTagRelations | PostWithUserTagRelatedPostRelations | Message,
     undefined,
-    { noOfRelatedPosts?: string }
+    { relatedPosts?: string }
   > = async (req, res) => {
     let post
     try {
       post = await this.postService.getSinglePost(
         req.params.id,
-        req.query.noOfRelatedPosts,
+        req.query.relatedPosts,
       )
     } catch (error) {
       logger.error({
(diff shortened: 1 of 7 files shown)


server/src/modules/post/post.controller.ts
```diff
@@ -159,17 +163,23 @@ export class PostController {
   /**
    * Get a single post and all the tags and users associated with it
    * @param postId Id of the post
+   * @query relatedPosts if true, return related posts
```
LoneRifle, line 166: this should be corrected to better reflect the parameter concerned

server/src/modules/post/post.service.ts
```diff
@@ -74,13 +78,55 @@ export class PostService {
...
+      ],
+      group: 'id',
+      order: [
+        [Sequelize.col('relatedTags'), 'DESC'],
+        ['views', 'DESC'],
+      ],
     })
-    return existingTags
+
+    return relatedPosts.slice(0, numberOfRelatedPosts)
```
LoneRifle: =/ consider the use of `limit` instead of slicing

server/src/modules/post/post.controller.ts
```diff
@@ -159,17 +163,23 @@ export class PostController {
...
    */
-  getSinglePost: ControllerHandler<{ id: string }, Post | Message> = async (
-    req,
-    res,
-  ) => {
+  getSinglePost: ControllerHandler<
+    { id: string },
+    PostWithUserTagRelations | PostWithUserTagRelatedPostRelations | Message,
+    undefined,
+    { noOfRelatedPosts?: string }
```
LoneRifle: I think we could get away with `relatedPosts` and make it a number, and it will be sensible enough. Do also consider adding the necessary validator to `post.routes`

LoneRifle (changes requested):
per individual comments

server/src/modules/post/post.service.ts
```diff
@@ -74,13 +78,55 @@ export class PostService {
...
+        include: [
+          [Sequelize.fn('COUNT', Sequelize.col('tags.id')), 'relatedTags'],
+        ],
+      },
+      where: {
+        id: {
+          [Op.ne]: post.id,
+        },
+      },
+      raw: true,
```
```suggestion
```

server/src/modules/post/post.controller.ts
nauynix: Added

LoneRifle (approved):
lgtm

PR #236, opened 2021-09-07
feat: preview image in image upload

### LINKED ISSUE #172: text overflow when inserting image
image

nauynix:
## Problem

The url link overflows when inserting image. It is also specific to the browsers chrome and safari, but wraps as expected in firefox.

image

- Closes #172

## Solution

Instead of trying to wrap the url, let the dialog show the preview of the images instead. Not only is the image constrained in the popup, it is also a better visual indicator to the user that the image upload worked. Tested to work on safari, firefox and chrome (on MacOS).

https://github.com/jpuri/react-draft-wysiwyg/issues/346

## After

https://user-images.githubusercontent.com/20250559/132284592-4fb6d142-c63c-4a87-ac09-0a4fedbf5824.mov


LoneRifle (approved):
lgtm

PR #237, opened 2021-09-07
feat(auth): make login otp email informational

### LINKED ISSUE #39: Informative OTP email
Currently our otp email is very bare. Can follow FormSG's email:
Screenshot 2021-08-05 at 2 49 44 PM

LinHuiqing:
- add more general information i.e. where email is coming from
- add IP address where login request is coming from

closes #39

## Problem

Closes #39.

## Solution

**Features**:

- add client machine's IP address information which sent tried to login

**Improvements**:

- add information on what OTP is for
- add fallback instructions if OTP does not work

## Before & After Screenshots

**BEFORE**:
Screenshot 2021-09-07 at 4 44 11 PM

**AFTER**:
Screenshot 2021-09-07 at 8 55 59 PM

## Tests

- (In test suite) `sendLoginOtp` in `src/modules/mail/__tests__/mail.service.spec.ts`.
- When user enters their email into the form (at https://ask.gov.sg/login), they should receive an email similar to what is seen in the after screenshot.
  - User should be able to login by entering the OTP in the login form.
  - User should be able to match the IP address to their own IP address, or find the rough location where the IP is from on the internet.

### DIFF
--- a/server/src/modules/auth/auth.controller.ts
+++ b/server/src/modules/auth/auth.controller.ts
@@ -111,10 +111,7 @@ export class AuthController {
 
     const otp = generateRandomDigits(OTP_LENGTH)
     const hashedOtp = await hashData(otp)
-    const ip =
-      ((req.headers['x-forwarded-for'] as string) || '').split(',')[0] ||
-      req.socket.remoteAddress ||
-      ''
+    const ip = req.header('CF-Connecting-IP') || req.ip
     try {
       await Token.destroy({ where: { contact: email } })
       await Token.create({ contact: email, hashedOtp })
--- a/server/src/modules/auth/auth.controller.ts
+++ b/server/src/modules/auth/auth.controller.ts
@@ -111,10 +111,14 @@ export class AuthController {
 
     const otp = generateRandomDigits(OTP_LENGTH)
     const hashedOtp = await hashData(otp)
+    const ip =
+      ((req.headers['x-forwarded-for'] as string) || '').split(',')[0] ||
+      req.socket.remoteAddress ||
+      ''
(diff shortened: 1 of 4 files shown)


LoneRifle:
Please have a look at https://github.com/opengovsg/checkfirst/blob/develop/src/server/utils/express.ts to figure out how we should be getting the IP of the calling client when the request is coming through Cloudflare

server/src/modules/auth/auth.controller.ts
```diff
@@ -111,10 +111,11 @@ export class AuthController {
 
     const otp = generateRandomDigits(OTP_LENGTH)
     const hashedOtp = await hashData(otp)
+    const ip = req.header('CF-Connecting-IP') || req.ip
```
LoneRifle, line 114: Maybe leave commentary here explaining what this is doing, and why. Take a cue from the comments in the code from where you found this line.

LoneRifle (approved):
lgtm otherwise

server/src/modules/auth/auth.controller.ts
LoneRifle, line 114: So ie...
```suggestion
    // On staging and production, we use CloudFlare which adds a CF-Connecting-IP
    // header with every request which contains only the origin IP
    // See https://support.cloudflare.com/hc/en-us/articles/200170986-How-does-CloudFlare-handle-HTTP-Request-headers-
    // As a fallback, use req.ip, automatically parsed from the x-forwarded-for
    // header if we configure app.set('trust proxy').
    const ip = req.header('CF-Connecting-IP') || req.ip
```

PR #242, opened 2021-09-08
docs(auth): provide and edit jsdocs for AuthService

LoneRifle:
## Context

Copyedit existing jsdocs for AuthService, and provide some for `createToken()`


nauynix (approved):
lgtm

PR #248, opened 2021-09-08
feat: instrument with Datadog

LoneRifle:
## Problem

Monitoring should be moved to a centralised location. We should also find a better
tool than Cloudwatch.

## Solution

Use Datadog as our monitoring tool of choice, exporting sys and app
metrics as well as logs either via Datadog Agent or directly.

- use the following tags as recognised by Datadog:
  - `service:askgovsg`
  - `env:staging` or `env:production`
- create two new env vars - `DD_API_KEY` and `DATADOG_ENV`
  - note that `DATADOG_ENV` is distinct from `NODE_ENV`;
    the latter has no notion of staging
- add .ebextensions script to install Datadog agent
- ensure inclusion in `deploy.zip` sent to Elastic Beanstalk
- add connect-datadog as Express middleware
- add datadog-winston for exporting logs to Datadog

References:
https://docs.datadoghq.com/integrations/amazon_elasticbeanstalk/?tab=singlecontainer#alternate-datadog-agent-configuration

## Deploy Notes

As noted in solution. Notably, .ebextensions script introduced to install and configure Datadog Agent

- `DD_API_KEY` : Datadog API Key for AskGov
- `DATADOG_ENV`: distinct from `NODE_ENV` (which only specifies build environment). Possible values are `staging` or `production`

**New dependencies**:

- `hot-shots` : client for StatsD/Datadog Agent
- `connect-datadog`: integrations for Connect/Express
- `datadog-winston`: winston transport, exporting logs via RESTful API


LoneRifle:
We might use `hostname_file` to obtain instance-id for the agent once DataDog/datadog-agent#8684 is released via v7.31.0

Gregcake:
### RE @TODO
There's a specific order of how the hostname is determined. Instance-id is likely being submitted (should be able to find it somewhere on the UI), but that value is not guaranteed to be used as the instance hostname.

https://docs.datadoghq.com/agent/faq/how-datadog-agent-determines-the-hostname/?tab=agentv6v7#potential-host-names

There are a number of params in datadog.yaml that you can pass to set the hostname as well.
https://github.com/DataDog/datadog-agent/blob/f2a92c38dd6c8852996e9432d0bb7bec221baa16/pkg/config/config_template.yaml#L59-L89

LoneRifle:
> ### RE @todo
> There's a specific order of how the hostname is determined. Instance-id is likely being submitted (should be able to find it somewhere on the UI), but that value is not guaranteed to be used as the instance hostname.
> 
> https://docs.datadoghq.com/agent/faq/how-datadog-agent-determines-the-hostname/?tab=agentv6v7#potential-host-names
> 
> There are a number of params in datadog.yaml that you can pass to set the hostname as well.
> https://github.com/DataDog/datadog-agent/blob/f2a92c38dd6c8852996e9432d0bb7bec221baa16/pkg/config/config_template.yaml#L59-L89

I got there in the end. For some reason in Elastic Beanstalk datadog-agent is unable to query the metadata service, so I decided to go with the approach per changes

server/src/bootstrap/config/datadog.ts
```diff
@@ -0,0 +1,60 @@
+import convict, { Schema } from 'convict'
+import { readFileSync } from 'fs'
+
+function getInstanceId() {
+  let instanceId = 'i-unknown'
+  try {
+    // Find the instance id as described in files commonly
+    // found in cloud-init hosts, including EC2 instances
+    instanceId = readFileSync('/var/lib/cloud/data/instance-id', 'utf-8')
```
mantariksh, line 9: I think the canonical way to do this is to query the [metadata endpoint](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html), but that would make this whole process async, which would be Very Annoying

server/src/bootstrap/config/datadog.ts
```diff
@@ -0,0 +1,60 @@
...
+  } catch (error) {
+    // Do nothing and carry on,
+    // unable to log error
```
mantariksh: should we default to `console.error` just so we can debug if it actually fails?

.ebextensions/datadog.config
```diff
@@ -0,0 +1,76 @@
+option_settings:
```
mantariksh, line 3: possible to add the source in a comment in this file itself pls, so we don't have to refer back to the PR to know where this came from?

mantariksh (approved):

server/src/bootstrap/config/datadog.ts
LoneRifle: aye, indeed

liangyuanruo: @LoneRifle FYI this try..catch..finally was flagged up by [this linting rule](https://eslint.org/docs/rules/no-unsafe-finally) when I copied it to GoGovSG

server/src/bootstrap/index.ts
```diff
@@ -131,6 +136,30 @@ const apiOptions = {
   enquiries: new EnquiryController({ enquiryService, recaptchaService }),
 }
 
+const moduleLogger = createLogger(module)
+
+if (baseConfig.nodeEnv === Environment.Prod) {
+  app.use(
+    connectDatadog({
+      response_code: true,
+      tags: [`service:${datadogConfig.service}`, `env:${datadogConfig.env}`],
```
timotheeg, line 145: These tags are probably not needed here, since they are set at agent-level

Gregcake: You'll probably want to pass in the [unified service tags](https://docs.datadoghq.com/getting_started/tagging/unified_service_tagging?tab=kubernetes) (env, service, version) as environment variables, not set them at the agent-level (imagine if you run multiple apps/services on a single EC2, can't assign tags to specific services at the agent level).

Full list of env variables can be found [here](https://docs.datadoghq.com/tracing/setup_overview/setup/nodejs?tab=containers#tagging).

Example:
```
DD_ENV="prod"
DD_SERVICE="example"
DD_VERSION="1.0.1"
DD_LOGS_INJECTION=true
node my_app.js
```

LoneRifle: Those are actually being set using env vars =) given that we only run one application per EC2 instance (elastic beanstalk after all), this should be okay.

LoneRifle: I think given our context, either should be okay. It might work cleaner if we set the tags explicitly in the application, but I don't foresee us worrying too much about tag organisation given that for us, a service will constitute an entire product in OGP.

PR #275, opened 2021-09-13
feat: replace authentication using JWT with session cookies 

### LINKED ISSUE #250: replace JWT with session tokens
JWT for authentication has several issues, highlighted http://cryto.net/~joepie91/blog/2016/06/13/stop-using-jwt-for-sessions/. Replace JWT with session tokens generated by server using `express-session`, and still using the `token` table.

nauynix:
## Problem

The server currently sends a JSON Web Token (JWT) to the client after authentication so that the client does not have to keep authenticating. This opens up to two issues. The first is that the token will be stored in the browser's localStorage and may only be retrieved via javascript, so any attacker supplied JS that passes the Content Security Policy can access it. Secondly, there is no way to invalidate a token if it was found out to be compromised or for any other reason, other than to change our JWT secret, which invalidates all users. Hence it is preferable to generate a session cookie to send to each individual user, and if they are authenticated, the session is stored in a new table 'Sessions'.

Authentication middleware was also not consistently used, with some `controllers` opting to check authentication themselves instead of using the middleware.

- Closes #250

## Solution

Use `Express-session` to manage sending and retrieving session cookies, 
`Passport` for authentication middleware, `Passport-local` to authenticate via email and otp, `Connect-session-sequelize` to store session in database via sequelize. For testing of authentication flow, `Supertest-session` is used to store and send cookies in tests.

The authentication middleware checks if a user is authenticated, and if it is not, sends back a `401` without moving on to the next controller. All routes that require authentication will use the middleware, and `controllers` are not responsible for checking authentication and can assume that `req.user.id` exists.

Security features:

- Since session tokens are verified by secret, generate session token only after logging in so an attacker w/o credentials cannot get any verified unauthenticated session token easily to mitigate session fixation.
- Regenerate session token after logging in, even if user has a valid session token.

Details ...

- Remove storing of JWT in client and `useLocalStorage` since the browser will automatically store cookies.
- Logging out now calls a new `/auth/logout` endpoint to tell the server to delete the session from the database.
- Sequelize script to create `session` table. The columns were choosen by following the defaults generated by `connect-session-sequelize`
- Logic for verifying otp is shifted from `auth.controller` to `passport/localStrategy`
- Typically, `passport.serializeUser` and `passport.deserializeUser` will convert between an id and user. However, as the previous JWT only stores the id, most of the services will take in `user.id` for querying. Hence I made the decision to have `deserializeUser` not convert id back to user.
- Add dependency injection where needed to aid in testing.
- `Controllers` and `services` that grab userId from `x-auth-token` Token will now instead take the userId directly from `req.user.id` after authentication.
- New file `express-auth.ts` modified from FromSG to set up any integrated tests in the future that require authentication, used here for `auth.routes`.
- Sessions are now valid for 24 hours (previously 1 hour). After which they will be deleted from the server, and any user that logs in with that session will not be authenticated.
- New `/server/tests/` folder to store helper test files that will not be compiled into build.

## Before & After Screenshots

**AFTER**:

https://user-images.githubusercontent.com/20250559/133567531-70db8d48-1db1-45bf-bf0e-464e0b1186b5.mov

## Deploy Notes

Create the new `session` table before deploying by running the migration script.
Add the `SESSION_SECRET` env var
Since this will log out any users still using JWT, this change should be pushed only when there are few users online.

- `SESSION_SECRET` : the secret to generate sessions
- `JWT_SECRET`: Deprecated, remove in future PR

**New scripts**:

- `20210909124846-create-sessions.js` : create session table.

**New dependencies**:

`Express-session` to manage sending and retrieving session cookies
`Passport` for authentication
`Passport-local` to allow for authentication via username and password
`Connect-session-sequelize` to store session in database via sequelize

**Dev dependencies**
`Supertest-session` to test authentication and storing session cookies


client/src/contexts/AuthContext.tsx
```diff
@@ -40,33 +35,27 @@ export const AuthProvider = ({
         }
       })
       .catch((reason: AxiosError) => {
-        // Catch 401 which signals an unauthorized user, which is not an issue
-        if (!(reason.response && reason.response.status === 401)) {
+        // Catch 400 or 401 which signals an unauthorized user, which is not an issue
```
mantariksh: why would a 400 signal an unauthorized user? doesn't look like that status is returned by this endpoint, and if we're returning it then it doesn't seem like the right status code

client/src/contexts/AuthContext.tsx
```diff
@@ -40,33 +35,27 @@ export const AuthProvider = ({
...
+        if (
+          !(
+            reason.response &&
```
mantariksh: do we still need this truthiness check? and if so can we just do `reason.response?.status` below?

client/src/contexts/AuthContext.tsx
```diff
@@ -40,33 +35,27 @@ export const AuthProvider = ({
...
-    onSuccess: ({ token: receivedToken }) => {
-      setToken(receivedToken)
-      setApiClientToken(receivedToken)
+    onSuccess: () => {
       whoami()
     },
   })
 
   const logout = () => {
+    ApiClient.get('/auth/logout')
```
mantariksh: what happens if the async call fails? do we need some error handling here?

server/src/bootstrap/passport/local.strategy.ts
```diff
@@ -0,0 +1,41 @@
+import passport from 'passport'
```
mantariksh, line 1: does the filename need to be dot-separated? would `localStrategy.ts` make more sense as a filename? `.strategy.ts` doesn't seem to be a reusable extension, unlike something like `.controller.ts`

server/src/modules/auth/auth.routes.ts
```diff
@@ -65,5 +65,13 @@ export const routeAuth = ({
     controller.handleSendLoginOtp,
   )
 
+  /**
+   * Logout
+   * @route   POST /api/auth/logout
+   * @returns 200 if logged out
+   * @access  private
+   */
+  router.get('/logout', controller.handleLogout)
```
mantariksh: don't think this should be a GET request, doesn't it modify the `sessions` table and hence change database state?

mantariksh (commented):
some prelim comments

server/src/bootstrap/passport/passport.ts
```diff
@@ -0,0 +1,27 @@
+import express from 'express'
```
mantariksh: instead of making consumers import from `passport/passport`, why not name this `index.ts`?

server/src/bootstrap/passport/passport.ts
```diff
@@ -0,0 +1,27 @@
...
+import passport from 'passport'
+import { localStrategy } from './local.strategy'
+import { Token, User } from '../../models'
+import { ModelCtor } from 'sequelize/types'
+
+const passportConfig = (
```
mantariksh: nit but prefer named exports because [reasons](https://basarat.gitbook.io/typescript/main-1/defaultisbad)

client/src/contexts/AuthContext.tsx
nauynix: image
You are right that this branch will not throw 400. The current develop branch will throw 400 if no auth. I am thinking of accepting both status first, then finally change to accept 401 only when the session is fully merged.

client/src/contexts/AuthContext.tsx
nauynix: Good catch

server/src/bootstrap/passport/local.strategy.ts
nauynix: Passport has multiple authentication strategy, and convention is each strategy will reside in a `xxx.strategy.ts` file eg. `facebook.strategy.ts`. Given that we will only use local in the foreseeable future, I think either way is fine.

server/src/modules/auth/auth.routes.ts
nauynix: Followed forms on this one. Happy to change to delete if you still think that is the way to go.

 https://github.com/opengovsg/FormSG/blob/232ff3ec94a090c3854e5b96a71c4d3f6dd036fe/src/app/routes/api/v3/auth/auth.routes.ts#L91

server/src/bootstrap/passport/passport.ts
nauynix: Thanks, will use this convention in the future.

client/src/contexts/AuthContext.tsx
nauynix: Scratch that. I will update it to only accept 401 because we will be pushing this change when there are few users anyways.

server/src/modules/auth/auth.routes.ts
nauynix: You are right, `POST` should be used instead (not `DELETE` as there is no resource to identify).

client/src/contexts/AuthContext.tsx
mantariksh: the new client will only ever see the new server and hence will never see 400s, so there's no point supporting both 400 and 401. lmk if that makes sense

server/src/util/express-auth.ts
```diff
@@ -0,0 +1,65 @@
...
+    (cookie) => cookie.name === 'connect.sid',
+  )
+  expect(sessionCookie).toBeDefined()
+
+  // Clear this test's mocked spies so calls do not pollute calling test.
+  // Note that the spies are still mocking the functions, so the onus is on the
+  // calling test to restore or implement new mocks when needed.
+  otpSpy.mockClear()
+  return request
+}
```
LoneRifle: Why is this in source and not in `__tests__`?

server/src/bootstrap/passport/local.strategy.ts
LoneRifle: side note - this strategy constitutes biz logic, so I expect tests for this

LoneRifle (changes requested):
lgtm otherwise, see comments

server/src/util/express-auth.ts
nauynix: Good catch. Should be in `__tests__` so it does not get picked up by tsc.

nauynix: On that note, `jest-db.ts` should be in `__tests__` too as its only used in tests. Shall we create a seperate test folder under `server/tests` for these helper files? I am hesitant to create a new folder for one or two files, but I think there will be more helper tests files in the future that should be consolidated somewhere.

LoneRifle: `server/tests` sounds good for now, go ahead!

server/src/bootstrap/passport/local.strategy.ts
nauynix: Added tests

LoneRifle (approved):

LoneRifle:
Merge when ready

Issue #218, opened 2021-09-06
Shared API types between frontend and backend


mantariksh:
APIs are a contract between the frontend and backend. This means that the client and server should agree on the shape of every API. In order to ensure that the frontend and backend agree on the API shapes without duplicating code, it makes sense for API types to be shared between the frontend and backend.

Implementing this will involve the following steps:
1. Edit the build pipeline to accommodate shared code between the `client` and `server` directories.
2. Create base types for each database table and have the model types inherit from the shared base types.
3. Create types for each API's return shape by composing the base types, and use these types in both the frontend services (which call the API) and the backend controllers (which serve the responses).

PR #287, opened 2021-09-14
fix: convert all id types to number

mantariksh:
## Problem

`id`s are typed as `string` throughout the codebase, but in reality they are of type `number` at runtime. I verified this on the client by checking the types of API responses as well as on the server by logging `typeof` in several services for several model IDs (`post`, `answer`, `tag` etc). I also verified that in the authentication JWT, the type of the user ID is also `number`.

Prerequisite for #218 

## Solution

Convert all `id` types to `number`. Note that for React Router, `match.params.id` is of type `string` at runtime and this is already typed correctly in the code, so I left this alone.

### Type of `req.params.id`

One wrinkle was that Express always types route parameters as `string`. This means that in the controller layer, `id`s are strings when they are derived from the route parameters, but they are numbers in the model layer. This inconsistency might cause confusion for future developers.

I decided to type all IDs in the **service layer and below** as `number`. This made sense to me because IDs should be numbers from the perspective of a service, whose job is to implement business logic and interface with the model layer, not to deal with HTTP requests. Since IDs being strings in the controller layer is a peculiarity of Express requests, it should be the controller's job to format the ID in a manner which the service can understand.

In turn, this brought up questions of type safety in the services, since calling the `Number` constructor on a non-number in the controller could result in `NaN` being passed to the service. Hence I introduced additional regex guards in the relevant routes to ensure that the `Number` constructor could be called safely in the controllers.

### DIFF
--- a/client/src/components/EditButton/EditButton.component.tsx
+++ b/client/src/components/EditButton/EditButton.component.tsx
@@ -20,7 +20,7 @@ import { ConfirmDialog } from '../ConfirmDialog/ConfirmDialog.component'
 import { useStyledToast } from '../StyledToast/StyledToast'
 
 interface EditButtonProps {
-  postId: string
+  postId: number
   onDeleteLink?: string
 }
 
--- a/client/src/services/AgencyService.ts
+++ b/client/src/services/AgencyService.ts
@@ -1,7 +1,7 @@
 import { ApiClient } from '../api'
 
 export type Agency = {
-  id: string
+  id: number
   email: string
   shortname: string
   longname: string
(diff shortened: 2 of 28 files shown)


LoneRifle:
express-validator should have validation and sanitization available to validate and coerce `req.params.id` into a Number. Perhaps this should be implemented as a separate PR.

LoneRifle (approved):
lgtm. Consider earlier comment about using express-validator to sanitize `req.params`

mantariksh:
@LoneRifle thanks, agree that this should be implemented separately so the implementation can deal with all APIs thoroughly. Tracked in #289.

PR #288, opened 2021-09-14
refactor: create Post and Agency shared types

mantariksh:
Creates a shared frontend/backend `Post` base type. Also combines `PostStatus` type which was previously separate but identical across the frontend and backend.

Part of #218 
Depends on #287


LoneRifle:
Interesting - why the need for a base collection of types? one assumes that the types found in `types/` are already base in nature

LoneRifle (approved):
I'm going to trust your judgment on `types/base` for now, and catch up later on. Thanks for doing this!

mantariksh:
>Interesting - why the need for a base collection of types? one assumes that the types found in types/ are already base in nature

the idea is that `types/base` will contain all the basic model types, while a future `types/api` will contain the API types, which will be compositions of the base types

PR #295, opened 2021-09-15
fix(searchbox): inject agencyShortName, don't derive from params

LoneRifle:
## Problem

When creating an enquiry from a question, autocomplete suggestions are not confined to the agency
that created the question.

## Solution

The agency short name may not always be available in path params,
so accept it as an injected property instead.

- Change SearchBox to accept `agencyShortName`, not `useParams()`
- Get HomePage to inject agency into CitizenRequest as-is, instead of
  using AskGov default, since SearchBox can cope with an undefined
  agency
  - Move AskGov defaults to CitizenRequest and EnquiryModal

## Screenshots

Note all suggestions being specific to Parking
image

## Tests

- Create (but don't submit) enquiry autosuggestions in:
  - AskGov landing page
  - Agency-specific page
  - Question
- For each scenario, verify that the suggestions are constrained to agency, or in the case of the AskGov landing page, span across agencies

### DIFF
--- a/client/src/components/CitizenRequest/CitizenRequest.component.tsx
+++ b/client/src/components/CitizenRequest/CitizenRequest.component.tsx
@@ -7,7 +7,7 @@ import { useStyledToast } from '../StyledToast/StyledToast'
 import { useGoogleAnalytics } from '../../contexts/googleAnalytics'
 import * as FullStory from '@fullstory/browser'
 
-const CitizenRequest = ({ agency }: { agency: Agency }): JSX.Element => {
+const CitizenRequest = ({ agency }: { agency?: Agency }): JSX.Element => {
   const toast = useStyledToast()
   const {
     onOpen: onEnquiryModalOpen,
@@ -15,14 +15,15 @@ const CitizenRequest = ({ agency }: { agency: Agency }): JSX.Element => {
     isOpen: isEnquiryModalOpen,
   } = useDisclosure()
   const googleAnalytics = useGoogleAnalytics()
+  const agencyName = agency?.shortname || 'AskGov'
 
   const sendOpenEnquiryEventToAnalytics = () => {
     googleAnalytics.sendUserEvent(
       googleAnalytics.GA_USER_EVENTS.OPEN_ENQUIRY,
-      agency.shortname,
+      agencyName,
     )
     FullStory.event(googleAnalytics.GA_USER_EVENTS.OPEN_ENQUIRY, {
-      enquiry_str: agency.shortname,
+      enquiry_str: agencyName,
     })
   }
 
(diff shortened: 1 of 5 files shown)


client/src/components/CitizenRequest/CitizenRequest.component.tsx
```diff
@@ -36,7 +37,7 @@ const CitizenRequest = ({ agency }: { agency: Agency }): JSX.Element => {
     captchaResponse: string,
   ): Promise<void> => {
     const mail: Mail = {
-      agencyId: agency.id ? [agency.id] : [],
+      agencyId: agency?.id ? [agency?.id] : [],
```
mantariksh, line 40: is there a possibility of the `id` being 0, which would make the boolean check unintentionally falsy?

mantariksh, line 40: is there a possibility of the `id` being 0, which would make the boolean check unintentionally falsy?

client/src/pages/SearchResults/SearchResults.component.jsx
```diff
@@ -53,7 +53,12 @@ const SearchResults = () => {
             <span style={{ color: '#acb2b8', fontSize: '12px' }}>
               Results for {searchQuery}
             </span>
-            <SearchBox placeholder={'Search...'} name={'search'} pt={'mt8'} />
+            <SearchBox
+              agencyShortName={agency}
```
mantariksh: nit: would prefer naming this variable `agencyShortName` so it's clearly not an object of type `Agency`

client/src/components/EnquiryModal/EnquiryModal.component.tsx
```diff
@@ -75,15 +75,16 @@ export const EnquiryModal = ({
         <ModalContent w="660px">
           <ModalHeader>
             <Text textStyle="h2" color="secondary.700">
-              {`${agency.shortname.toUpperCase()} Enquiry Form`}
+              {`${(agency?.shortname ?? '').toUpperCase()} Enquiry Form`}
```
mantariksh, line 78: will there be an extra weird space in front of "Enquiry Form" if `agency` is not defined?

LoneRifle: Browser happily renders the extra space away
image

client/src/components/CitizenRequest/CitizenRequest.component.tsx
LoneRifle: Good point! the side-effect is somewhat desired - Sequelize should not be generating agencies with an id of `0`.

mantariksh (approved):

PR #296, opened 2021-09-15
fix(auth): return error for auth attempts with invalid emails

mantariksh:
## Problem

Previously, if there was an existing user with an invalid public service
email, we would return undefined tokens and display names from the OTP
verification endpoint. While the likelihood of such a user existing is low,
this has the added effect of making it difficult to create a shared API type
for this endpoint between the frontend and backend.

Part of #218 

## Solution

This change removes the possibility of returning undefined tokens and
display names from the OTP verification endpoint by shifting the
isOfficerEmail check earlier in the pipeline and returning an error
early for invalid emails.


LoneRifle (approved):
lgtm

PR #297, opened 2021-09-15
refactor(shared): create remaining shared model types

mantariksh:
- Creates shared model types for `Answer`, `Tag` and `Permission`
- Creates shared enums for `TagType` (previously separate enums in frontend and backend) and `PermissionType` (previously hardcoded strings)


LinHuiqing (approved):
lgtm

PR #298, opened 2021-09-15
refactor: create shared API types for auth endpoints

mantariksh:
Creates shared types between frontend and backend for authentication endpoints.

Part of #218 
Depends on #297

### DIFF
--- a/shared/src/types/base/common.ts
+++ b/shared/src/types/base/common.ts
@@ -1,7 +1,6 @@
 import { z } from 'zod'
 
-export const BaseModel = z.object({
-  id: z.number(),
+export const SequelizeTimestamps = z.object({
   // createdAt and updatedAt are of type Date
   // in the backend, but are received by the frontend
   // as strings because
@@ -10,4 +9,8 @@ export const BaseModel = z.object({
   // on the string/Date
   createdAt: z.date().or(z.string()),
   updatedAt: z.date().or(z.string()),
+})
+
+export const BaseModel = SequelizeTimestamps.extend({
+  id: z.number(),
 })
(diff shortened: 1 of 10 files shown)


shared/types/base/permission.ts
```diff
@@ -1,15 +1,16 @@
...
+import { BaseModel } from './common'
 
 export enum PermissionType {
   Answerer = 'answerer',
   Admin = 'admin',
 }
 
-// no BaseModel as Permission table has no id
-export const Permission = z.object({
+// Permission table has no id
```
LoneRifle: If so.. would it be better to model the Sequelize timestamps as its own object, then have BaseModel inherit from it? this would allow you to avoid the `omit()` declaration

mantariksh: done in https://github.com/opengovsg/askgovsg/pull/298/commits/2c73056781865a0ff7a1d03189bc5e85a2839e2e

shared/src/types/base/common.ts
```diff
@@ -1,7 +1,6 @@
 import { z } from 'zod'
 
-export const BaseModel = z.object({
-  id: z.number(),
+export const SequelizeTimestamps = z.object({
```
LoneRifle: nit - I generally favour the Scala approach of naming partial classes (traits) as `HasXXX`. Only a suggestion though.

mantariksh: would this mean something like `HasTimestamps`? sorry not familiar with that convention

LoneRifle: Aye, something like `HasSequelizeTimestamps`, so that you get something like `HasSequelizeTimestamps.extend(...)`

mantariksh: done in https://github.com/opengovsg/askgovsg/pull/298/commits/b0104045582b52da668bb3b72acd845f5491f82f

LoneRifle (approved):
Lgtm, here's to a cleaner codebase!

PR #300, opened 2021-09-15
refactor(shared): move to shared/src, fix server build

LoneRifle:
## Context

Align shared project structure to the other TS projects, and fix build
for server so that the appropriate types are referenced at runtime.

- `shared/types` -> `shared/src/types`
  - Update `client/tsconfig.paths.json` to reflect change in path
- Introduce one extra step to build shared types before `npm run dev`
  and `npm run test`, so that shared types are available to server
  at runtime
- Configure `server/tsconfig.json` to understand built shared types
  - Alias `~shared` to `../shared/build`, for tsconfig.json and jest
  - Use `tsconfig-paths` package so that server understands aliases
  - Register `tsconfig-paths/register` at `npm run (dev|start)`
- Update server codebase to use `~shared` to avoid referencing `/build`


wanlingt (approved):

mantariksh (commented):
hmm this shouldn't be required in theory with the `--build` option passed to `tsc`

mantariksh:
still looking at the code but preliminary question - I'm guessing this means `ts-node-dev` doesn't restart when there are changes in `shared`? we might need to add options to `ts-node-dev` to get it to watch `shared`

LoneRifle:
> still looking at the code but preliminary question - I'm guessing this means `ts-node-dev` doesn't restart when there are changes in `shared`? we might need to add options to `ts-node-dev` to get it to watch `shared`

not entirely sure - I did register the module to resolve aliased paths, so it should know to rebuild in response to changes in `shared/`. I suppose we could put in a PR to correct this if I'm proven wrong

server/package.json
```diff
@@ -5,9 +5,9 @@
   "description": "The backend for Ask.gov.sg",
   "main": "./build/index.js",
   "scripts": {
-    "start": "node ./build/index.js",
+    "start": "node -r tsconfig-paths/register ./build/index.js",
     "build": "tsc -b",
-    "dev": "ts-node-dev --respawn --transpile-only --inspect -- ./src/index.ts",
+    "dev": "ts-node-dev -r tsconfig-paths/register --respawn --transpile-only --inspect -- ./src/index.ts",
```
mantariksh: tried it out, the TS recompiles but the server doesn't restart when changes are made to `shared`. easy fix:
```suggestion
    "dev": "ts-node-dev -r tsconfig-paths/register --respawn --transpile-only --inspect --watch ../shared/src -- ./src/index.ts",
```

mantariksh (approved):
thank you for this and sorry for not realising that I broke the server.

interestingly, in VSCode, clicking into `~shared` on the server still brings me to the TS file, not the compiled JS. this is because of the `references` in `tsconfig` (I know this because I  when I tried removing the `references` and clicking into `~shared`, it brought me to the declaration files).

PR #301, opened 2021-09-15
fix(build): move shared into server

mantariksh:
## Problem

The `shared` folder is not being referenced correctly in the built `server` files. This is because [TypeScript project references](https://www.typescriptlang.org/docs/handbook/project-references.html) only reference the declaration files of the listed projects, not actual built JavaScript. Hence the build initially succeeded because `shared` was only exporting types, but the moment it started attempting to export objects in #288, compilation broke.

## Solution

Move `shared` into `server/src`. This still makes semantic sense - shared types are inherited by the client from the server. This allows us to get around the build problems because the `server` simply gets built as before, and the client just aliases into `server/src/shared/*` instead of `shared/*`.

Another alternative would have been to move the `rootDir` of the server build from `server/src` into the root of the project. However, this would have resulted in a strange structure of `server/build`, which would have been separated into `server/build/server/src` and `server/build/shared`.


LoneRifle (changes requested):
Could we consider #300 instead?

mantariksh:
Closing in favour of #300

PR #311, opened 2021-09-16
refactor(spinner): migrate to Chakra and TypeScript

### LINKED ISSUE #273: Migrate Spinner to Chakra + TypeScript
As a gentle introduction to the frontend part of the codebase,
migrate the Spinner to use Chakra components and TypeScript

wanlingt:
- Convert Spinner component to use Chakra components and TypeScript
- AnswerSection component uses Chakra's SkeletonText directly

## Problem

Closes #273 

## Solution

Previously, 2 types of spinners were being used to display the loading state - one for pages and another for non-page sections (e.g. answer section, sending OTP)
These have been replaced with the Chakra spinner. For the answer section, Chakra's SkeletonText is used directly.

## Before & After Screenshots

**BEFORE**:

https://user-images.githubusercontent.com/56983748/133585462-4a283630-2281-4c9e-97e6-ce011d83c8aa.mov

https://user-images.githubusercontent.com/56983748/133585485-b1f53a1c-c96e-4c29-b612-508c72848352.mov

**AFTER**:

https://user-images.githubusercontent.com/56983748/133585788-20bd8abd-ff0d-4dd5-9727-63f169ed800d.mov

https://user-images.githubusercontent.com/56983748/133585621-fb1f08ea-fdca-48dc-a8a2-a7a684cbbe66.mov

## Tests

- Verify that pages use the updated Chakra spinners. These include
  - Landing page
  - Answer page
  - Header
  - Login page (sending OTP)
  - Edit form page
  - Post form page
  - Search results page


client/src/components/Spinner/Spinner.component.tsx
```diff
@@ -0,0 +1,22 @@
...
+  type?: string
+  width: string | number
+  height: string | number
+}
+
+const Spinner = ({ type, width, height }: SpinnerProps): JSX.Element => {
+  return type === 'paragraph' ? (
+    <SkeletonText />
+  ) : (
+    <div className="spinner" style={{ width: `${width}`, height: `${height}` }}>
```
nauynix: Since we are migrating to chakra, the `div` should also be chakra and the properties are styled in this file, and thus we can remove the corresponding `scss` file.

client/src/components/Spinner/Spinner.component.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
+import './Spinner.styles.scss'
+
+import { SkeletonText, Spinner as PageSpinner } from '@chakra-ui/react'
+
+interface SpinnerProps {
+  type?: string
```
mantariksh: this isn't just any `string` right - you have a specific set of strings which are allowed (e.g. `'paragraph'`)? if so we can type this more strongly with [string literal types](https://www.typescriptlang.org/docs/handbook/literal-types.html#string-literal-types)

client/src/components/Spinner/Spinner.component.tsx
```diff
@@ -0,0 +1,22 @@
...
+  width: string | number
+  height: string | number
```
mantariksh: noticed that these are required types, yet in `AnswerSection.component.jsx`, they are not provided (and the code doesn't complain because the component isn't in TypeScript). perhaps these should be optional?

alternatively, consider this - it seems your goal here is to provide the consumer with two options:
1. render a `SkeletonText`, in which case styles need not be provided
2. render a `PageSpinner`, in which case the width and height must be provided

this approach allows you to specify the type more explicitly:
```ts
type SpinnerProps = 
  | { type: 'paragraph' }
  | { type: 'page'; width: string | number; height: string | number }
```

however, notice that the `type` key is redundant here. you can tell what the type is from the presence/absence of the width and height keys. this allows us to simplify further:
```ts
type SpinnerProps = 
  | Record<string, never>
  | { width: string | number; height: string | number }
```
where `Record<string, never>` indicates the absence of any props, and instead of `type === 'paragraph'`, you can check `width !== undefined`.

I know this probably seems SUPER nitpicky but a) hopefully it's useful for learning and b) strong type systems are built from solid basic types!

client/src/components/Spinner/Spinner.component.tsx
mantariksh: instead of wrapping in an extra `div` just for styling, why not just use the ChakraUI styling options?
```tsx
<PageSpinner w={width} h={height} />
```

wrapping in an extra div could also be slightly annoying for styling maintainability since the `Spinner` component returns two possible things with two different levels of DOM nesting

mantariksh (commented):
comments mostly for learning! not requesting changes bc mostly nitpicks, totally okay if this goes in as-is.

for your consideration - the `Spinner` component doesn't seem to be adding much abstraction over the ChakraUI `SkeletonText` and `PageSpinner` components. in fact, there are loads of props which consumers might want to pass to `SkeletonText` and `PageSpinner` (`width`, `height`, `speed`, `thickness` etc) but this component makes that impossible because it does not pass on any props to the ChakraUI components.

two possible solutions:
1. get rid of this component and just use `SkeletonText` and `PageSpinner` directly everywhere
2. make `SpinnerProps` extend from ChakraUI's `SpinnerProps` or `SkeletonProps` (not sure how easy this would be since `SpinnerProps` and `SkeletonProps` are different), and pass on the props internally to the relevant component

mantariksh:
also since our users are primarily on mobile, can we make it  a habit to include screenshots from mobile pls?

client/src/components/Spinner/Spinner.component.tsx
nauynix: I believe we still have to wrap it in a `Center` or equivalent to center it.

wanlingt: Yeah the alignment didn't work without `Center`, so I've left it in

client/src/components/Spinner/Spinner.component.tsx
wanlingt: I've gone with a combination of the 2 suggestions: 
1. Used SpinnerText directly in AnswerSection component
2. make `SpinnerProps` extend from ChakraUI's `SpinnerProps`

wanlingt:
> also since our users are primarily on mobile, can we make it a habit to include screenshots from mobile pls?

added!

wanlingt:
> also since our users are primarily on mobile, can we make it a habit to include screenshots from mobile pls?

Added!

client/src/components/Spinner/Spinner.component.tsx
```diff
@@ -0,0 +1,38 @@
+import { Center, Spinner as PageSpinner } from '@chakra-ui/react'
+
+interface SpinnerProps {
```
mantariksh: actually ChakraUI exports all component props, which you can see from the [source](https://github.com/chakra-ui/chakra-ui/blob/main/packages/spinner/src/spinner.tsx), so you can just import `SpinnerProps`

client/src/components/Spinner/Spinner.component.tsx
```diff
@@ -0,0 +1,38 @@
+import { Center, Spinner as PageSpinner } from '@chakra-ui/react'
```
mantariksh: nit: the name `PageSpinner` doesn't make its distinction from `Spinner` clear. I'd suggest `ChakraSpinner` to make it clear that `Spinner` is wrapping a ChakraUI component.

client/src/components/Spinner/Spinner.component.tsx
```diff
@@ -0,0 +1,38 @@
...
+  width,
+  height,
+  size,
+  thickness,
+  emptyColor,
+  color,
+  speed,
+}: PageSpinnerProps): JSX.Element => {
+  return (
+    <Center className="spinner" h={height} w={width}>
```
mantariksh: would it work if we applied the width and height on the `Spinner` itself instead? that would allow you to spread the props so the whole component can look like this:
```ts
import {
  Center,
  Spinner as ChakraSpinner,
  SpinnerProps,
} from '@chakra-ui/react'

const Spinner = (props: SpinnerProps): JSX.Element => {
  return (
    <Center>
      <ChakraSpinner {...props} />
    </Center>
  )
}

export default Spinner
```

client/src/components/RichText/LinkControl.tsx
```diff
@@ -90,7 +90,11 @@ export const LinkControl = ({
   return (
     <div className={styles.linkControl}>
       <div onClick={signalExpandShowModal} className="rdw-option-wrapper">
-        <BiLink size="18px" />
+        <Tooltip label="Link" placement="bottom" openDelay={500} fontSize="xs">
+          <span>
+            <BiLink size="18px" />
+          </span>
+        </Tooltip>
```
mantariksh: is this change meant to be in this PR? usually we try to scope PRs to a single set of changes, for a couple of reasons:
1. easier to review
2. if we need to revert commits, we can do so in isolation without affecting other changes

client/src/components/Spinner/Spinner.component.tsx
wanlingt: Looks this if we apply `width` and `height` on `Spinner`, so probably not the best.

https://user-images.githubusercontent.com/56983748/133736244-d8d9b44a-7202-488a-98e6-1dd89f8b5982.mov



Without `width` and `height` applied to `Center`:

Screenshot 2021-09-17 at 2 40 50 PM

With `height` applied to `Center`:

Screenshot 2021-09-17 at 2 41 03 PM

LinHuiqing: Not sure if this is a dumb question, but is there a need for a custom component file for the spinner? Would it make sense for us to use the Spinner component from Chakra directly? It looks very similar based on the docs: https://chakra-ui.com/docs/feedback/spinner

If we want customisations (which are shared across the site), we could extend the theme as well: https://chakra-ui.com/docs/theming/customize-theme#customizing-component-styles

mantariksh: is it just a matter of adjusting the width and height to a more appropriate size?

wanlingt: unfortunately no, that changes the size of the spinner but not its positioning

client/src/components/Spinner/Spinner.component.jsx
```diff
@@ -1,15 +0,0 @@
-import React from 'react'
-import './Spinner.styles.scss'
-
-import { ReactComponent as PageSpinner } from '../../assets/PageSpinner.svg'
-import { ReactComponent as ComponentSpinner } from '../../assets/three-dots.svg'
```
LoneRifle, line 5: you can take the opportunity to eliminate the two SVGs in this PR

LoneRifle (approved):
lgtm, respond to comments then merge when ready

mantariksh (approved):
nice, the `centerWidth` and `centerHeight` solution is good because it still allows the ChakraUI inbuilt `width` and `height` to be applied to the Spinner if necessary

client/src/components/Spinner/Spinner.component.jsx
wanlingt: deleted

PR #314, opened 2021-09-16
fix(test): ensure Jest types are imported

### LINKED ISSUE #313: Type definitions and namespace for jest no longer appears
When test files are specified not to be compiled as introduced in #286, the type definitions and namespace for jest no longer appears and the editor thinks jest variables are errors.

image

mantariksh:
## Problem

Closes #313

## Solution

Extract a separate `tsconfig.build.json` for building the backend. `tsconfig` applies to the whole project (including tests), whereas `tsconfig.build.json` applies only to build files.

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -6,7 +6,7 @@
   "main": "./build/index.js",
   "scripts": {
     "start": "node -r tsconfig-paths/register ./build/index.js",
-    "build": "tsc -b",
+    "build": "tsc -p tsconfig.build.json",
     "dev": "ts-node-dev -r tsconfig-paths/register --respawn --transpile-only --inspect --watch ../shared/src -- ./src/index.ts",
     "test": "jest",
     "lint": "eslint . --quiet --fix --ext '.ts'",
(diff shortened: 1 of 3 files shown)


nauynix (approved):
lgtm

mantariksh:
holding off from merging until I'm certain @LoneRifle won't kill me in my sleep

server/package.json
```diff
@@ -6,7 +6,7 @@
   "main": "./build/index.js",
   "scripts": {
     "start": "node -r tsconfig-paths/register ./build/index.js",
-    "build": "tsc -b",
+    "build": "tsc -p tsconfig.build.json",
```
LoneRifle, line 9: I assume we don't have to invoke `-b` anymore since we don't actually need the `references` bit?

LoneRifle (approved):
lgtm

server/package.json
mantariksh: yep so we're not using TS project references anymore. but interestingly in VSCode if you remove the `references` bit from `tsconfig`, clicking into a `~shared` import brings you to the compiled `.d.ts` file instead of the `.ts` file, which is annoying DX, which is why I haven't removed it. I recognise that's not the best reason to keep it around though, so am okay to remove in a separate PR as well.

PR #319, opened 2021-09-17
feat(funnels): add analytics events to track click / browse / search

### LINKED ISSUE #294: Add analytics events to track conversion funnel
# Problem

Currently we are limited in our ability to track how users interact with our site. To make better informed design decisions, we then need to gather data on how users are using the site.

# Solution

Add GA and FullStory events to capture when user uses the site, then create conversion funnels on FullStory. There are mainly 3 conversion funnels we are looking at to figure out how users find the answers to the questions they have (or don't):

- Search
- Click Tag
- Browse (i.e. scroll through questions on the home page)

Diagram of how this is currently being implemented:

AskGov Conversion Funnels

Dashboard for conversion funnels: https://app.fullstory.com/ui/12P870/dashboards/4538974704672768

LinHuiqing:
## Problem

Currently we are limited in our ability to track how users interact with our site. To make better informed design decisions, we then need to gather data on how users are using the site.

Closes #294. Refer to issue for more details.

## Solution

Add custom events to be fired for Google Analytics (GA) and FullStory (FS) to catch. FS is then configured to show the conversion funnels on the dashboard.

How events captured are combined to create conversion funnels:

AskGov Conversion Funnels

**Features**:

- Add 'Click Tag' events for both GA and FS. These events are tagged with the following details:
  - String of the tag which is clicked.
  - Time at which the tag is clicked starting from when app is loaded.
- Add 'Browse' events for both GA and FS.
  - Agency page on which browsing was done, identified by the agency's short name.
  - Percentage of the page which was scrolled.

**Improvements**:

- Refactor both `client/src/services/googleAnalytics.tsx` and `client/src/contexts/googleAnalytics.tsx` so that `googleAnalytics` can be used in TypeScript files.
- Refactor `TagBadge` component to TypeScript.

## Tests

With an IP address that is not blocked from recording by FS (refer to [this page](https://app.fullstory.com/ui/12P870/settings/recording)), navigate to the home page.

### Browse Events

1. Scroll a distance then navigate to another page (possibly by clicking into a question).
2. Close the page.
3. On FS, a new session should be registered. This session should include 'Browse' events which have the agency's short name for the agency page that is browsed and the percentage of the page which has been scrolled.
4. On GA, 'Browse' events should be captured on the Realtime dashboard.

### Click Tag Events

1. Click on tags, either those under questions or on the side panel.
2. Close the page.
3. On FS, a new session should be registered. This session should include 'Click Tag' events which have the names of the tags which are clicked on and the time after page load which the tag click was made.
4. On GA, 'Click Tag' events should be captured on the Realtime dashboard.


client/src/components/PostList/PostList.component.jsx
```diff
@@ -18,6 +19,54 @@ const PostList = ({ posts, defaultText, alertIfMoreThanDays, showViews }) => {
...
+  const googleAnalytics = useGoogleAnalytics()
+  const updateMaxScroll = () => {
+    googleAnalytics.maxScroll = Math.max(
+      googleAnalytics.maxScroll,
+      window.pageYOffset,
+    )
+  }
+
+  const sendMaxScrollToAnalytics = (maxScrollPossible) => {
+    maxScrollPossible = Math.max(maxScrollPossible, googleAnalytics.maxScroll) // this is required to restrict the percentage to 100
```
mantariksh: why would `googleAnalytics.maxScroll` ever exceed `maxScrollPossible`? wouldn't that just make `maxScrollPercentage` 1?

client/src/components/PostList/PostList.component.jsx
```diff
@@ -18,6 +19,54 @@ const PostList = ({ posts, defaultText, alertIfMoreThanDays, showViews }) => {
...
+  }
+
+  useEffect(() => {
+    resetMaxScrollAndAddListener()
+    const pageHeight = Math.max(
+      document.body.scrollHeight,
+      document.body.offsetHeight,
+      document.documentElement.clientHeight,
+      document.documentElement.scrollHeight,
+      document.documentElement.offsetHeight,
```
mantariksh, line 67: what's the difference between all of these? is there no single source of truth for the height of the viewport?

client/src/components/SearchBox/SearchBox.component.jsx
```diff
@@ -69,7 +69,7 @@ const SearchBox = ({
     FullStory.event('Search', {
       timeToFirstSearch_int: timeToFirstSearch,
     })
-    googleAnalytics.setIsFirstSearch(true)
+    googleAnalytics.isFirstSearch = true
```
mantariksh: given that `isFirstSearch` is state in the `GoogleAnalyticsProvider`, we should be using the `setIsFirstSearch` setter so that the change propagates correctly - see [useState docs](https://reactjs.org/docs/hooks-state.html)

client/src/components/TagBadge/TagBadge.component.tsx
```diff
@@ -0,0 +1,50 @@
...
+  tagName: string
+  tagType: TagType
+  agency: Agency
+}): JSX.Element => {
+  const TagComponent = <div className="tag-badge">{tagName.toUpperCase()}</div>
+
+  const googleAnalytics = useGoogleAnalytics()
+
+  const sendClickTagEventToAnalytics = () => {
+    const timeToTagClick = new Date().getTime() - googleAnalytics.appLoadTime
```
mantariksh: no need to change but just FYI these are equivalent
```suggestion
    const timeToTagClick = Date.now() - googleAnalytics.appLoadTime
```

client/src/components/TagBadge/TagBadge.component.tsx
```diff
@@ -0,0 +1,50 @@
...
+    googleAnalytics.sendUserEvent(
+      googleAnalytics.GA_USER_EVENTS.CLICK_TAG,
+      tagName,
+      timeToTagClick,
+    )
+    googleAnalytics.sendTiming(
+      'User',
```
mantariksh, line 30: noticed we are slotting this into the same category as `sendUserEvent` - would providing a different category make the GA console more organised? up to you, I'm not super familiar with this

mantariksh (commented):
for your consideration - what do you think about keeping the `maxScroll` state in `PostList.component` itself, perhaps via a [ref](https://reactjs.org/docs/hooks-reference.html#useref), instead of in the `googleAnalytics` service?

a couple of reasons for this:
1. single responsibility principle. imo the GA service should just be in charge of firing GA events, not keeping track of state in other components.
2. imo the GA service is already keeping track of too much state in other components, which makes things confusing. when reading the GA service in isolation, it is not at all obvious what `maxScroll` is doing. imagine if every single component which fired GA events started sharing state with the GA service - how would we know what each variable represents?

client/src/components/PostList/PostList.component.jsx
LinHuiqing: If I'm not wrong, there are some browser compatibility issues which cause the total scroll height (i.e. `maxScrollPossible`) to differ on different browsers even with the same screen size (refer to [this post](https://stackoverflow.com/questions/17688595/finding-the-maximum-scroll-position-of-a-page)). There is a better API to track this ([window.scrollMaxY](https://developer.mozilla.org/en-US/docs/Web/API/Window/scrollMaxY)), but it's only available on Firefox so...

When the `maxScroll` exceeds `maxScrollPossible`, its not a lot usually within 20%. Since we're just using this as a gauge to see whether users browse, I think the precision doesn't have to be very high?

client/src/components/PostList/PostList.component.jsx
LinHuiqing: Sadly, there isn't :( Unless we assume that users only use certain browsers. Refer to [this post](https://stackoverflow.com/questions/17688595/finding-the-maximum-scroll-position-of-a-page).

client/src/components/SearchBox/SearchBox.component.jsx
LinHuiqing: I changed it to a variable, because I'm not very convinced of the use of states when the value is only changed once and is only read after.

client/src/components/TagBadge/TagBadge.component.tsx
LinHuiqing: I think its fine because we can search by the event label on GA? Either way should be fine, but User is a pretty big category so I just assumed it belonged there.

LinHuiqing:
> for your consideration - what do you think about keeping the `maxScroll` state in `PostList.component` itself, perhaps via a [ref](https://reactjs.org/docs/hooks-reference.html#useref), instead of in the `googleAnalytics` service?
> 
> a couple of reasons for this:
> 
> 1. single responsibility principle. imo the GA service should just be in charge of firing GA events, not keeping track of state in other components.
> 2. imo the GA service is already keeping track of too much state in other components, which makes things confusing. when reading the GA service in isolation, it is not at all obvious what `maxScroll` is doing. imagine if every single component which fired GA events started sharing state with the GA service - how would we know what each variable represents?

For this, I started with the approach which you suggested actually, but I changed to the approach I am using now because googleAnalytics was storing the appLoadTime (so I assume that was the preferred approach). Otherwise, the appLoadTime could have been a ref too I think?

Anyway, yea I agree its quite bloated and should be in the component itself as it does not need to be shared.

client/src/components/TagBadge/TagBadge.component.tsx
LinHuiqing: Nice, looks cleaner, thanks! Will change it together with other places where this is done.

mantariksh:
>Otherwise, the appLoadTime could have been a ref too I think?

agreed, although I think `appLoadTime` is a bit more forgivable since it's a genuinely app-wide variable, whereas `maxScroll` is actually specific to `PostList`

client/src/components/PostList/PostList.component.jsx
mantariksh: I see! can we leave some documentation here (can link to the relevant SO post also) for future developers?

client/src/components/PostList/PostList.component.jsx
mantariksh: thanks! can link to the SO post in a comment in the code pls?

client/src/components/PostList/PostList.component.jsx
```diff
@@ -18,6 +19,55 @@ const PostList = ({ posts, defaultText, alertIfMoreThanDays, showViews }) => {
...
+    )
+    const maxScrollPossible = pageHeight - window.innerHeight
+    return () => removeListenerAndSendEvent(maxScrollPossible)
+  }, [])
```
LoneRifle: Some commentary of what this does would be helpful. It's not immediately obvious that the hook installs a callback that gets triggered only when the component unloads

LoneRifle (approved):
lgtm otherwise, nicely done!

Issue #271, opened 2021-09-13
Provide unit test coverage for `env/` module 


LoneRifle:
As a gentle introduction to testing and the server codebase,
provide test coverage for either or both the abovementioned parts.

nauynix:
`createOfficer` is not being used as users have to have their emails be in the database already. Consider removing it.

PR #329, opened 2021-09-20
test(user): add unit test for UserService

wanlingt:
## Problem

No tests exist for User Service. Additionally, createOfficer will never be used as user emails already exist in the database.

## Solution

- provide unit tests for UserService (re #271 )
- remove createOfficer from UserService

### DIFF
--- a/server/src/modules/user/__tests__/user.service.spec.ts
+++ b/server/src/modules/user/__tests__/user.service.spec.ts
@@ -1,0 +1,35 @@
+import { UserService } from '../user.service'
+
+describe('UserService', () => {
+  const userService = new UserService()
+
+  afterEach(async () => {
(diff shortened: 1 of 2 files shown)


LoneRifle (changes requested):
`createOfficer` is actually used here:
https://github.com/opengovsg/askgovsg/blob/048f4aad2b0c675a74c7ef923758836f91789ac7/server/src/modules/auth/auth.controller.ts#L195

nauynix:
You can remove `createOfficer` now that it is no longer used by `auth`

LoneRifle:
Let's revisit this some other time. This was mostly an exercise to have you more familiar with the codebase - I think it's clear you're quite comfortable now.

PR #341, opened 2021-09-21
refactor(agency): use shared types, dep injection for service

LoneRifle:
## Problem

The agency module currently does not have test coverage and does not use 
shared types. Its service also does not use dep injection for its models.

## Solution

- Create a custom definition for sequelize models since standard ones
  are not quite accurate. More precisely, a Model<M> will have
  properties of M as well as Model, but this is not declared so.
- Rework AgencyService so that the Sequelize model is injected rather
  than imported. Take care to use the new imported types to avoid
  returning Sequelize objects to callers
- Adjust AgencyController and bootstrap to account for changes
- Provide test coverage for the module
- Take the opportunity to only log failure to get instance-id in prod

### DIFF
--- a/server/src/bootstrap/config/datadog.ts
+++ b/server/src/bootstrap/config/datadog.ts
@@ -1,5 +1,6 @@
 import convict, { Schema } from 'convict'
 import { readFileSync } from 'fs'
+import { baseConfig, Environment } from './base'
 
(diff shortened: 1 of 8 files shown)


nauynix (approved):
lgtm

PR #357, opened 2021-09-23
fix(bootstrap): send index.html with express.static

LoneRifle:
## Problem

The backend currently serves index.html using `Response.sendFile()`, 
which incurs a file read with each request. Due to lack of rate-limiting,
this exposes us to a possible DoS by sending multiple requests to 
overwhelm the filesystem.

## Solution

Avoid reading the file repeatedly by using express.static to read the
file just once at runtime, and serve it from in-memory


mantariksh (approved):
LG<sup>TM</sup>

PR #358, opened 2021-09-23
feat(breakpoints): add breakpoints for devices

LinHuiqing:
Added breakpoints for devices so specifications for mobile, table and desktop can be simply added
with: `{ base: <specs for mobile>, tablet: <specs for tablet>, desktop: <specs for desktop> }`

## Problem

The granular breakpoints for `xs`, `sm`, `md`, `lg` and `xl` are not very meaningful when development is centred around devices instead of those specific sizes.

## Solution

Add breakpoint specifications for `tablet` and `desktop` so development centred around devices will be easier and more readable.

To use code for specifications for different devices, simply use the following: `{ base: <specs for mobile>, tablet: <specs for tablet>, desktop: <specs for desktop> }`. For usage examples: https://chakra-ui.com/docs/features/responsive-styles

## Tests

Change display size and check if content on display is expected.

### DIFF
--- a/client/src/theme/breakpoints.ts
+++ b/client/src/theme/breakpoints.ts
@@ -1,7 +1,13 @@
+// Note the following breakpoints for different devices:
+// - mobile: 479px and below
+// - tablet: 480-1439px
+// - desktop: 1440px and above
 export const breakpoints = {
   xs: '360px',
   sm: '480px',
   md: '768px',
   lg: '1024px',
   xl: '1440px',
+  tablet: '480px', // should be same as sm
+  desktop: '1440px', // should be same as xl
 }


client/src/theme/breakpoints.ts
```diff
@@ -1,7 +1,13 @@
...
+// - desktop: 1440px and above
 export const breakpoints = {
   xs: '360px',
   sm: '480px',
   md: '768px',
   lg: '1024px',
   xl: '1440px',
+  tablet: '480px', // should be same as sm
+  desktop: '1440px', // should be same as xl
 }
```
LoneRifle, line 13: if the devices are still coupled to the sizes, perhaps the following would allow for easier maintenance and self-documentation:

```ts
const sizeBasedBreakpoints = {
  xs: '360px',
  sm: '480px',
  md: '768px',
  lg: '1024px',
  xl: '1440px',
}

export const breakpoints = {
  ...sizeBasedBreakpoints,
  tablet: sizeBasedBreakpoints.sm,
  desktop: sizeBasedBreakpoints.xl,
}

**Review by LoneRifle — COMMENTED** [review]
I'm not convinced, for a couple of reasons:
- I don't remember the wider industry using breakpoint names other than the size ones found here. Do you have references from elsewhere where `tablet` and `desktop` breakpoints are used?
- Devices even of the same class can vary (see eg, an iPad mini vs iPad vs iPad Pro). I realise we may lack the bandwidth to cater to this many sizes now, but I would argue that aligning to what's more widely accepted would give us the option to have more sizes later, at the expense of remembering that our favoured breakpoints are `sm` and `xl` for now.

File: client/src/theme/breakpoints.ts (hunk 1 above)
Comment (LinHuiqing) [reply client/src/theme/breakpoints.ts:13]: Hm I think either way works? I did so because I recall using a check for device when using Vue in the past instead of pixel sizes and thought this would be easier for development as it reduces cognitive load. In the iPad / iPad Mini / iPad Pro use case where we need more granularity, I think we can just use the normal breakpoints then?

Comment (LinHuiqing) [reply client/src/theme/breakpoints.ts:13]: For reference Vue's UI library, Vuetify, allows users to easily check if the device is mobile through its breakpoint service. For more information: https://vuetifyjs.com/en/features/breakpoints/#breakpoint-service-object

Comment (LoneRifle) [reply client/src/theme/breakpoints.ts:13]: I see, so Vue takes this approach. Do you see equivalent practice in React?

Comment (LinHuiqing) [reply client/src/theme/breakpoints.ts:13]: Kind of? There are packages like [react-device-detect](https://github.com/duskload/react-device-detect) which help with this. There are 1.4k stars and 82 forks on the repository so its relatively popular as well.

Comment (LoneRifle) [reply client/src/theme/breakpoints.ts:13]: Tangential at best. @karrui - any opinions on breakpoint labels?

Comment (karrui) [reply client/src/theme/breakpoints.ts:13]: I'd say stick with sizing; xs -> xxl sizes? This allows [`chakra#useBreakpointValue`](https://chakra-ui.com/docs/hooks/use-breakpoint-value) or the [responsive syntax](https://chakra-ui.com/docs/features/responsive-styles) to be more intuitive imho. Since breakpoints are sorted in ascending values, if there are two of the same sizes i.e. `sizes.sm` and `sizes.tablet` for example, this may break how the responsive syntax is parsed.

Comment (LinHuiqing) [reply client/src/theme/breakpoints.ts:13]: Ah, I guess that's true if we use the array way. Okay then, let's just close this PR.

Comment (LinHuiqing) [reply client/src/theme/breakpoints.ts:13]: To add on, I'll just some comments about the breakpoints for mobile/ tablet/ desktop as they are the breakpoints that we are considering in the design system.

Comment (karrui) [reply client/src/theme/breakpoints.ts:13]: nonetheless, if you are only centered around the few breakpoints and if you really want to only call the breakpoints via their synthetic name, you can always declare the breakpoints (and override the defaults like this:)

```tsx
createBreakpoints({
  'base': '0px',
  'mobile': '?px',
  'tablet': '?px',
  'desktop': '?px',
  // Default breakpoints all "unused" with this hack
  'md': '9999px',
  'sm': '9999px',
  'lg': '9999px',
  'xl': '9999px',
  '2xl': '9999px',
});
```

this will allow the usage of both array and object responsive syntax, though now the developers has to make sure not to use any of the "default" breakpoints as reference (or their styles won't get applied until 9999px)

LinHuiqing: Hm, I don't think it's a good idea to make the default breakpoints useless in the case where we need finer granularity in the future (as mentioned by @LoneRifle). Thanks for the 'hack' though!

PR #375, opened 2021-09-27
fix(breakpoints): specify in em, not px

LoneRifle:
## Problem

Breakpoints specified in px are poorly handled by Chakra, which is unable to reconcile them against its own default breakpoints, specified in em. This triggers chakra-ui/chakra-ui#3042, and causes dependents (notably the boolean flag that chooses which pagination widget to use) to oscillate.

## Solution

Port opengovsg/formsg@54b9175 to avoid the abovementioned bug

#### Before

https://user-images.githubusercontent.com/10572368/134842304-3c409cfb-3756-4953-a043-86284bfed76b.mov

#### After

https://user-images.githubusercontent.com/10572368/134842273-f8e13101-d221-4d8e-b8fa-405a7a9f2014.mov

### DIFF
--- a/client/src/theme/breakpoints.ts
+++ b/client/src/theme/breakpoints.ts
@@ -1,7 +1,9 @@
-export const breakpoints = {
+import { createBreakpoints } from '@chakra-ui/theme-tools'
+
+export const breakpoints = createBreakpoints({
   xs: '22.5em',
   sm: '30em',
   md: '48em',
   lg: '64em',
   xl: '90em',
-}
+})
--- a/client/src/theme/breakpoints.ts
+++ b/client/src/theme/breakpoints.ts
@@ -1,7 +1,7 @@
 export const breakpoints = {
-  xs: '360px',
-  sm: '480px',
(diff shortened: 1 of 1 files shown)


client/src/theme/breakpoints.ts
```diff
@@ -1,7 +1,7 @@
 export const breakpoints = {
```
mantariksh: use [`createBreakpoints`](https://chakra-ui.com/docs/theming/theme#breakpoints)?

LinHuiqing (commented):
Other than @mantariksh's comment, lgtm

client/src/theme/breakpoints.ts
LinHuiqing: Sorry, this is my bad, I didn't use `createBreakpoints` initially. Would be a good idea based on the docs haha

LinHuiqing: I can help fix this?

mantariksh (approved):
LG<sup>TM</sup>

PR #388, opened 2021-09-28
Feat/neverthrow agency

arshadali172:
## Problem

Use of `try...catch` for error handling is [not type-safe](https://www.codementor.io/@supermacro/type-safe-error-handling-in-typescript-1bp40rs502). Hence we need to gradually migrate to [neverthrow](https://github.com/supermacro/neverthrow#top-level-api).

## Solution

- Introduce `neverthrow` to the agency module
- Introduce `ApplicationError` class and extend it to create appropriate error classes
- Fix tests with the use of `_unsafeUnwrap` and `_unsafeUnwrapErr`

## Tests

Current test suite should pass

## Deploy Notes

None


LoneRifle (approved):
lgtm

PR #389, opened 2021-09-28
feat(enquiries): allow agencies to refuse, providing message instead

LoneRifle:
## Problem and Solution
Some agencies are unable to handle enquiries coming in from AskGov, so provide them with the means to direct them elsewhere

- add `noEnquiriesMessage` to agencies shared type
  - define corresponding column in sequelize, migration script
- have AgencyService use shared type instead of defining its own
- if `noEnquiriesMessage` defined, present a simple text modal bearing
  the message instead of the EnquiryModal


nauynix (approved):
lgtm

PR #405, opened 2021-09-28
feat: add topics model to database schema

### LINKED ISSUE #396: Update database schema with Categories model
## Problem
The current database schema does not allow for posts to be organised by categories

## Solution
- Create Categories model
- Link each post to a category

wanlingt:
## Problem

The current database schema does not allow for posts to be organised according to agency-specific topics (formerly referred to as 'categories')

Closes #396 

## Solution

- Add a new model `Topic`
- Modify the `Post` model such that posts can be organised into topics. `topicId` has been added as a FK

Details ...
- Each topic knows its immediate parent; child topics carry a foreign key (`parentId`) to their parent topic. For parent topics, the `parentId` will be null.
- In the `posts` table, `topicId` is nullable. This means that `topics` are optional for each post.

## Tests
- After running the migration script, verify that the ERD diagram is updated with the new `Topic` model

## Deploy Notes

Create new `topics` table and update the `posts` table before deploying by running the new migration scripts.

**New scripts**:

- `20211010101959-create-topics.js` : create topics table
- `20211011014741-add-topic-to-posts.js` : add topicId as a new column in posts table


server/migrations/20210922101959-create-categories.js
```diff
@@ -0,0 +1,49 @@
...
+      },
+      parentId: {
+        allowNull: true,
+        type: Sequelize.INTEGER,
+        references: {
+          model: 'categories',
+          key: 'id',
+        },
+        onUpdate: 'CASCADE',
+        onDelete: 'SET NULL', //Cascade?
```
nauynix: Have you resolved if this should be set null or cascade?

server/migrations/20210923014741-add-category-to-posts.js
```diff
@@ -0,0 +1,20 @@
...
+        model: 'categories',
+        key: 'id',
+      },
+      onUpdate: 'CASCADE',
+      onDelete: 'SET NULL',
+    })
+  },
+
+  down: async (queryInterface) => {
+    await queryInterface.removeColumn('posts', 'agencyId')
```
nauynix: is `agencyId` supposed to be `categoryId` instead?

shared/src/types/base/category.ts
```diff
@@ -0,0 +1,10 @@
+import { z } from 'zod'
+import { BaseModel } from './common'
+
+export const Category = BaseModel.extend({
+  catname: z.string(),
+  parentId: z.string(),
```
nauynix: Is the parentId supposed to be a number?

server/migrations/20210923014741-add-category-to-posts.js
wanlingt: yes my bad

shared/src/types/base/category.ts
wanlingt: yup thanks for spotting!

server/migrations/20210922101959-create-categories.js
LoneRifle: should be `SET NULL`. I don't think it's a good idea if the parent gets deleted, leaving orphans in its wake.

server/migrations/20210922101959-create-categories.js
```diff
@@ -0,0 +1,49 @@
...
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    return queryInterface.createTable('categories', {
+      id: {
+        autoIncrement: true,
+        primaryKey: true,
+        type: Sequelize.INTEGER,
+      },
+      catname: {
```
LoneRifle:
```suggestion
      name: {
```
and elsewhere

server/seeders/20210928063247-was-add-categories.js
```diff
@@ -0,0 +1,57 @@
+'use strict'
```
LoneRifle: I think let's not worry about having a seeder script for this for now. Having chatted with others and seeing how AskGov has been most recently used, I have a sense that we should make _both_ categories _and_ tags optional. This would grant the agency some flexibility over how their information should be organised. They could either:
- Use categories only to organise information in traditional manner;
- Use tags only for a more ad-hoc approach that allows them to group questions under multiple (possibly cross-agency) topics, or;
- Use both approaches.

server/src/models/categories.model.ts
```diff
@@ -0,0 +1,26 @@
...
+import { Agency } from './agencies.model'
+
+export interface Category extends Model, CategoryBaseDto {}
+
+//constructor
+export const defineCategory = (
+  sequelize: Sequelize,
+  { Agency }: { Agency: ModelCtor<Agency> },
+): ModelCtor<Category> => {
+  const Category: ModelCtor<Category> = sequelize.define('category', {
```
LoneRifle: Consider typing it as `ModelDef<Category>` (the shared type you have defined). This allows you to avoid having to explicitly define the Category in line 5.

shared/src/types/base/category.ts
```diff
@@ -0,0 +1,10 @@
+import { z } from 'zod'
+import { BaseModel } from './common'
+
+export const Category = BaseModel.extend({
+  catname: z.string(),
```
LoneRifle:
```suggestion
  name: z.string(),
```

server/src/models/categories.model.ts
```diff
@@ -0,0 +1,26 @@
...
+    catname: {
```
LoneRifle:
```suggestion
    name: {
```

server/migrations/20210922101959-create-categories.js
```diff
@@ -0,0 +1,49 @@
...
+      createdAt: {
+        allowNull: false,
+        type: Sequelize.DATE,
+      },
+      updatedAt: {
+        allowNull: false,
+        type: Sequelize.DATE,
+      },
+      agencyId: {
+        allowNull: true, //to change to false once we implement this across all agencies
```
LoneRifle:
```suggestion
        allowNull: false
```
I think this should be false from the get-go, you can't really have categories without an agency. An agency however can get away without having categories, particularly since posts already have a nullable categoryId.

LoneRifle (changes requested):
I realise I'm setting a poor example by reviewing this PR at this hour. Nonetheless, do take a look at my initial comments.

server/migrations/20210922101959-create-categories.js
```diff
@@ -0,0 +1,49 @@
...
+      id: {
+        autoIncrement: true,
+        primaryKey: true,
+        type: Sequelize.INTEGER,
+      },
+      catname: {
+        type: Sequelize.TEXT,
+        allowNull: false,
+      },
+      createdAt: {
```
LoneRifle: We probably need a nullable description here.

LoneRifle:
actually, shall we rename this to topics instead, given that's what the public knows them by?

wanlingt:
> actually, shall we rename this to topics instead, given that's what the public knows them by?

can

server/seeders/20210928063247-was-add-categories.js
wanlingt: removed seeder script

LoneRifle (approved):

PR #413, opened 2021-09-29
feat(seo): update page titles according to agency

### LINKED ISSUE #401: Update page titles on agency pages
## Problem

Currently, the page titles for agency pages are still "AskGov" which is not only confusing for the user who only cares about the agency and not AskGov (sad), but discourages search engines from ranking us higher.

## Solution

Dynamic page titles which change according to agency which the AskGov site is for.

LinHuiqing:
- dynamically set HomePage and Post page's title when viewing agency

re #401

## Problem

Page title not being reflective of the agency which the page is serving.

Closes #401

## Solution

Dynamically render page title in home and post page components depending on agency.

## Before & After Screenshots

**BEFORE**:

Page title on agency (WAS) home page:

Screenshot 2021-09-29 at 8 38 18 AM

Page title on agency (WAS) post page:

Screenshot 2021-09-29 at 8 38 05 AM

**AFTER**:

Page title on agency (WAS) home page:

Screenshot 2021-09-29 at 8 34 43 AM

Page title on agency (WAS) post page:

Screenshot 2021-09-29 at 8 35 57 AM

## Tests

1. Navigate to an agency home page and check if the page title includes the agency name.
2. Navigate to an agency post page and check if the page title includes the agency name.

### DIFF
--- a/client/src/pages/HomePage/HomePage.component.jsx
+++ b/client/src/pages/HomePage/HomePage.component.jsx
@@ -11,12 +11,13 @@ import {
   Text,
 } from '@chakra-ui/react'
 import { useEffect, useState } from 'react'
-import { BiChevronUp, BiChevronDown } from 'react-icons/bi'
+import { BiChevronDown, BiChevronUp } from 'react-icons/bi'
 import { useQuery } from 'react-query'
 import { useLocation, useParams } from 'react-router-dom'
 import AgencyLogo from '../../components/AgencyLogo/AgencyLogo.component'
 import CitizenRequest from '../../components/CitizenRequest/CitizenRequest.component'
 import OfficerDashboardComponent from '../../components/OfficerDashboard/OfficerDashboard.component'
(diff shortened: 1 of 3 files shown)


wanlingt (approved):
lgtm

LoneRifle:
@LinHuiqing as a follow up to this, could we perhaps consider how we can set the title page _in the template_, ie _before_ React initialises? Would there be any meaningful improvement to our ranking as a result of that?

Bear in mind that this would mean how non-API paths are served by the backend.

LinHuiqing:
> @LinHuiqing as a follow up to this, could we perhaps consider how we can set the title page _in the template_, ie _before_ React initialises? Would there be any meaningful improvement to our ranking as a result of that?
> 
> Bear in mind that this would mean how non-API paths are served by the backend.

It seems that Google is able to index the question pages with the title though it is populated dynamically:

Screenshot 2021-09-30 at 2 00 40 PM

Will hold off the suggestion quoted for now.

PR #414, opened 2021-09-29
fix(seo): fix title in the case of an undefined agency

LinHuiqing:
## Problem

"undefined" was shown as agency name on the page title when the main AskGov page is viewed.

## Solution

Check if agency is defined. If so, update page title to include agency name. If not, use default title by setting `title` property of `PageTitle` as `undefined`.

## Before & After Screenshots

**BEFORE**:

image

**AFTER**:

image

## Tests

1. Navigate to the main AskGov page and check the page title. It should be the default page title.
2. Navigate to the agency main pages and check the page title. It should be in the format of "<agency name> FAQ - AskGov".

### DIFF
--- a/client/src/pages/HomePage/HomePage.component.jsx
+++ b/client/src/pages/HomePage/HomePage.component.jsx
@@ -62,7 +62,11 @@ const HomePage = ({ match }) => {
 
   return (
     <Flex direction="column" height="100%" className="home-page">
-      <PageTitle title={`${agency?.shortname.toUpperCase()} FAQ - AskGov`} />
+      <PageTitle
+        title={
+          agency ? `${agency?.shortname.toUpperCase()} FAQ - AskGov` : undefined
+        }
+      />
       <Box
         bg="primary.500"
         h={


client/src/pages/HomePage/HomePage.component.jsx
```diff
@@ -62,7 +62,11 @@ const HomePage = ({ match }) => {
 
   return (
     <Flex direction="column" height="100%" className="home-page">
-      <PageTitle title={`${agency?.shortname.toUpperCase()} FAQ - AskGov`} />
+      <PageTitle
```
nauynix, line 65: nit: does it make sense to conditionally render it then?

LinHuiqing: 1. Does nit mean new in town?
2. No, I don't think so. We still need the page title to display the default titles set.

LoneRifle: nit means nitpick =)

LoneRifle (approved):

PR #417, opened 2021-09-29
feat(post): reinstate last updated date

LoneRifle:
## Problem

The last updated date in Post does not reflect the last time content
was updated, as Post.views is updated whenever someone views the post.
It was thus not returned by the backend, with the created timestamp being
not as useful and thus hidden

## Solution
- replace moment with date-fns-tz in /client
- avoid updatedAt changes on post view
- expose updatedAt to frontend
- use date-fns-tz to display the date after the answer section

## Screenshot
image

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -6785,6 +6785,16 @@
         "whatwg-url": "^8.0.0"
       }
     },
+    "date-fns": {
+      "version": "2.24.0",
+      "resolved": "https://registry.npmjs.org/date-fns/-/date-fns-2.24.0.tgz",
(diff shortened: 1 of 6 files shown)


nauynix:
`UpdateAt` is being updated whenever someone views a post because it is incrementing the views count by one. However, it is possible to set the `silent` flag to prevent updating of updatedAt timestamp in a query.

https://sequelize.org/master/class/lib/model.js~Model.html#static-method-update

LoneRifle:
> `UpdateAt` is being updated whenever someone views a post because it is incrementing the views count by one. However, it is possible to set the `silent` flag to prevent updating of updatedAt timestamp in a query.
> 
> https://sequelize.org/master/class/lib/model.js~Model.html#static-method-update

good point! we still might have problems where viewers trying to change Post.views might be conflicting with other users trying to change the contents of the post, but I don't think this is a large problem for now

nauynix (approved):
lgtm

PR #418, opened 2021-09-29
feat: restrict number of attempts before deleting token

nauynix:
## Problem and solution

Currently login allows unlimited tries with the same hashed token, which is susceptible to brute force. Restrict the number of tries to 3 before deleting token, forcing user to generate a new one.

- Take the opportunity to convert `AuthForm` to typescript, add form validation where possible
- Change text to make it clearer to user to grab OTP from email

**AFTER**:

https://user-images.githubusercontent.com/20250559/135200634-a3622a80-ac4f-40b4-a94d-03c11739f4d0.mov

## Deploy Notes

Run migration script to add new `attempts` column to `Token`

### DIFF
--- a/server/src/bootstrap/passport/local.strategy.ts
+++ b/server/src/bootstrap/passport/local.strategy.ts
@@ -5,6 +5,8 @@ import { Token, User } from '../../models'
 import { verifyHash } from '../../util/hash'
 const LocalStrategy = passportLocal.Strategy
 
+export const MAX_OTP_ATTEMPTS = 3
+
 export const localStrategy = (
   Token: ModelCtor<Token>,
   User: ModelCtor<User>,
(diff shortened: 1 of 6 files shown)


server/src/bootstrap/passport/local.strategy.ts
```diff
@@ -20,9 +22,18 @@ export const localStrategy = (
...
+              return done(null, false, {
+                message:
+                  'You have hit the max number of attempts. Please request for a new OTP.',
+              })
+            } else {
+              await token.increment('attempts')
+              return done(null, false, {
+                message: 'OTP is invalid. Please try again.',
+              })
+            }
```
LoneRifle, line 36: This doesn't really stop a determined attacker though - asking for a new OTP is cheap. you have to figure a better way to deter the attacker, maybe with rate-limiting

LoneRifle (approved):
lgtm actually, thanks for spotting.

PR #421, opened 2021-09-29
feat: show link to external agency website

### LINKED ISSUE #415: Link out to main agency site on the navbar

nauynix:
## Problem

- Closes #415

## Solution

Add new column `website` to `agency`. Websites should contain the full path of the url, including protocol (http), otherwise the link will still stay in askgov. The button will display only the hostname instead of the full url to maintain legitimacy (`.gov`) and brevity.

If the website is empty or at homepage, the link does not show up.

## Before & After Screenshots

**BEFORE**:
Screenshot 2021-09-29 at 2 31 30 PM

**AFTER**:

https://user-images.githubusercontent.com/20250559/135215101-127181b1-7669-4d40-8445-ed924abde53a.mov

** Deploy notes **
Run the migration script to add new column

### DIFF
--- a/client/src/components/Header/Header.component.jsx
+++ b/client/src/components/Header/Header.component.jsx
@@ -60,17 +60,21 @@ const Header = () => {
     </Flex>
   )
 
-  const websiteLinks = (
-    <Link href={agency?.website} isExternal>
-      <Button
-        rightIcon={<BiLinkExternal color="white" />}
-        variant="link"
-        color="white"
-      >
-        Go to {agency?.shortname.toUpperCase()}
-      </Button>
-    </Link>
-  )
+  const websiteLinks = () => {
+    // Extract hostname from URL
+    const hostname = new URL(agency?.website).hostname
+    return (
+      <Link href={agency?.website} isExternal>
+        <Button
+          rightIcon={<BiLinkExternal color="white" />}
+          variant="link"
+          color="white"
+        >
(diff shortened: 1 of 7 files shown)


LinHuiqing (commented):
Shouldn't this be optimised for mobile (portrait) as well? As shown here: https://www.figma.com/file/O6lxjjfn04Ow6nn2TMStpk/AskGov-Design-Master-v2?node-id=5517%3A28555

client/src/components/Header/Header.component.jsx
```diff
@@ -51,6 +60,22 @@ const Header = () => {
...
+    // Extract hostname from URL
+    const hostname = new URL(agency?.website).hostname
+    return (
+      <Link href={agency?.website} isExternal>
+        <Button
+          rightIcon={<BiLinkExternal color="white" />}
+          variant="link"
+          color="white"
+        >
+          Go to {hostname}
```
LoneRifle, line 73: rather than this, could we just have a simple "Go to site"? @khaleedah wdyt?

LoneRifle (approved):
> Shouldn't this be optimised for mobile (portrait) as well? As shown here: https://www.figma.com/file/O6lxjjfn04Ow6nn2TMStpk/AskGov-Design-Master-v2?node-id=5517%3A28555

This will be approved for now, so that we can move on

PR #423, opened 2021-09-29
feat(seo): set existing images to lazy load

### LINKED ISSUE #391: Images should be lazy loaded
## Problem

Images are currently eager loads even though they may not be seen by the user. This causes page load times to be unnecessarily long. Longer page load time leads to lower search rankings as crawlers deem the page as a bad quality one.

## Solution

Set image elements to lazy load.

wanlingt:
## Problem

Images are currently eager loads even though they may not be seen by the user. This causes page load times to be unnecessarily long. Longer page load time leads to lower search rankings as crawlers deem the page as a bad quality one.

Closes #391 

## Solution

- Migrate AgencyLogo component to use Chakra so that we can utilise Chakra Image's loading prop
- Set loading to lazy for each existing image

## Before & After Screenshots
**BEFORE**:
Screenshot 2021-09-29 at 4 56 49 PM

**AFTER**:
135235566-62b32fa7-5a8e-48b4-a1d7-80035051150a

## Tests
- In each agency's homepage, check that the agency image displays correctly
- Check that the other images (lion's head masthead icon, footer, user avatar) display correctly


LinHuiqing (commented):
Just something I realised: these image optimisations are not applied to images which are uploaded as part of answers to questions. Considering our current pages has limited images from agencies, we are fine for now? Not sure if this will change with `LWC19` though...

LinHuiqing (approved):
Otherwise, edits lgtm

wanlingt:
> Just something I realised: these image optimisations are not applied to images which are uploaded as part of answers to questions. Considering our current pages has limited images from agencies, we are fine for now? Not sure if this will change with `LWC19` though...

Yeah, quite unfortunately it's not possible with the [current rich text editor we're using](https://www.npmjs.com/package/react-draft-wysiwyg). There are basically 2 parts to it:
1. the content we enter into the Answers section (in Draft.js format) is converted to HTML
    - Kinda hacky but we can add `loading="lazy"` to the `img` tag here by doing a replace on the html string
2. the HTML is converted back to Draft before being rendered on our Answer page
   - However `loading="lazy"` is stripped out when html is converted back to Draft, because [the library is not able to handle custom elements](https://github.com/jpuri/html-to-draftjs/issues/62) :(

We could look into exploring other options in the future, explanation above for posterity! Agree that it's probably not needed for now given the limited number of images per page.

PR #428, opened 2021-09-30
build(shared): ensure a clean build for every invocation

LoneRifle:
## Problem

`npm run build-shared` invokes `tsc` as-is, which exits early if it discovers `tsconfig.tsbuildinfo`, even if files are missing. This interferes with running in dev

## Solution

Ensure a clean build for every invocation


LinHuiqing (approved):
lgtm

PR #443, opened 2021-10-01
feat(rte): enforce alt text for images

LoneRifle:
## Problem

To ensure agencies maintain good SEO practices, images should have alt text annotations in questions and answers

## Solution

Enforce alt text for images in RichTextEditor

image


LinHuiqing (approved):
lgtm

PR #454, opened 2021-10-02
fix(search-results): add spacing between results, enquiry button

LoneRifle:
Self-explanatory, per image:
image


wanlingt (approved):
tested, lgtm

PR #457, opened 2021-10-03
fix(tag-panel): update tag panel copy for better cohesion

LinHuiqing:
- update home page title text when filtered by topic from "Questions tagged:" to "Questions related to:"
- organise imports in TagPanel

## Problem

> feedback from a user that said it wasn't clear that the topics on the left are directly related/ directly changes the questions tagged on the right (might be because the tags look different from the topic or something)

## Solution

Change home page title text from "Questions tagged:" to "Questions related to:" so users can more directly relate the panel to the tags clicked.

**Improvements**:

- organise imports for TagPanel

## Before & After Screenshots

**BEFORE**:

image

**AFTER**:

image

## Tests

Click on a topic either on the side panel or on a question. The home page's title text should be changed to "Questions related to: \<topic\>".

### DIFF
--- a/client/src/components/TagPanel/TagPanel.component.tsx
+++ b/client/src/components/TagPanel/TagPanel.component.tsx
@@ -10,9 +10,12 @@ import {
   Text,
   VStack,
 } from '@chakra-ui/react'
+import * as FullStory from '@fullstory/browser'
 import { ReactElement } from 'react'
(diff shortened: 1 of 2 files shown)


LoneRifle (approved):
I remember that we chose topics over tags as the former term was more easily understood by the public. Let's go with tags for now and see what happens.

Schedule a release as needed

Issue #394, opened 2021-09-28
Populate sensible `<meta>` tags in questions and question lists


LoneRifle:
To allow crawlers to better understand and present AskGov information to users, embed `<meta>` tags and/or Open Graph tags using information that can be quickly obtained from the db.

You may wish to use Facebook's Sharing Debugger and similar tools to help determine effectiveness of your implementation.

LinHuiqing:
What do you mean by question lists here actually? What should the `<meta>` be for that?

LoneRifle:
I refer to the agency-specific pages, apologies for not being clearer

Issue #256, opened 2021-09-09
HTTP status code should be 404 for non-existent paths


LoneRifle:
**Describe the bug**
If a non-existent page is visited, the service returns 200 OK rather than 404, even though the page does not actually exist.

**To Reproduce**
Visit ask.gov.sg/does-not-exist - observe that the service returns 200 OK even though the page indicates Not Found

**Expected behavior**
HTTP status should be 404

LinHuiqing:
Fixed by PR https://github.com/opengovsg/askgovsg/pull/494.

PR #494, opened 2021-10-05
feat(social-media-share): serve different static index.html files based on path

### LINKED ISSUE #361: Optimise agency pages for Search Engine Optimisation (SEO)
## Problem

The site currently ranks very low on Google search. This means that the visibility of the site is severely compromised and many citizens are likely to try looking up the questions they have on Google.

## Solution

Optimise agency pages and question pages to improve Google search rankings.

(Details on the specifics of the solution(s) will be added as more research as done.)

### LINKED ISSUE #403: Explore react rendering options for better SEO
## Problem 

Search crawlers need HTML to understand what the page is doing. However, as we are using React to build the page, the crawler will first see Javascript code, which it then has to send to the renderer to render recursively until either 1) site is fully rendered or 2) crawl budget is exceeded. The latter may mean that some of our pages are not indexed and thus not searchable.

## Solution

Explore rendering options other than the default client-side rendering (which I assume we are using). These options include:

- Client-side rendering with bootstrapped data
- Server-side rendering to static content
- Server-side rendering with rehydration
- Pre-rendering to static content
- Pre-rendering with rehydration

There is also isomorphic react, where we can consider using Next.js.

Helpful link: https://www.toptal.com/react/react-seo-best-practices

LinHuiqing:
## Problem

Social media crawlers are only able to see initial files served statically by server (i.e. `index.html`). To make our website more shareable on social media, these initial files should be populated with titles, meta description and open graph tags.

Related to #394 

## Solution

Get `server/src/bootstrap/index.ts` to modify `index.html` before serving them.

**Features**:

- (server/src/bootstrap/index.ts) send modified index.html for /agency/:shortname and /questions/:id paths for production

**Improvements**:

- (client/public/index.html) added og tags

## Before & After Screenshots

**BEFORE**:

Social media link preview:

Screenshot 2021-10-05 at 5 07 44 PM

**AFTER**:

Main AskGov page:

Screenshot 2021-10-05 at 5 13 29 PM

Agency-specific page:

Screenshot 2021-10-05 at 5 13 56 PM

Post page:

Screenshot 2021-10-05 at 5 14 58 PM

## Tests

Use [Facebook Sharing Debugger](https://developers.facebook.com/tools/debug/) to check that the title, meta description and og tags are populated. Specifically, test this for the following pages:

1. Main AskGov page
2. Agency-specific page
3. Post page

## Deploy Notes

**New dependencies**:

- `dependency` on `server`: 
  - `sanitize-html`
  - `cheerio`

**New dev dependencies**:

- `dependency` on `server`:  `@types/sanitize-html`

### DIFF
--- a/server/src/bootstrap/index.ts
+++ b/server/src/bootstrap/index.ts
@@ -62,6 +62,8 @@ import { datadogConfig } from './config/datadog'
 import { passportConfig } from './passport'
 import sessionMiddleware from './session'
 
+import cheerio from 'cheerio'
+
 export { sequelize } from './sequelize'
 export const app = express()
 
@@ -114,13 +116,14 @@ const mailService = new MailService({
 const postService = new PostService({ Answer, Post, PostTag, Tag, User })
 const enquiryService = new EnquiryService({ Agency, mailService })
 const recaptchaService = new RecaptchaService({ axios, ...recaptchaConfig })
+const answersService = new AnswersService()
 
 const apiOptions = {
   agency: new AgencyController({ agencyService }),
   answers: {
     controller: new AnswersController({
       authService,
-      answersService: new AnswersService(),
+      answersService: answersService,
     }),
     authMiddleware,
   },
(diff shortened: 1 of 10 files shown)


LinHuiqing:
To be honest, `og:url` is still missing for some reason :/ not quite sure why or how to debug that?

server/src/bootstrap/index.ts
```diff
@@ -194,9 +198,100 @@ if (baseConfig.nodeEnv === Environment.Prod) {
...
+          'content',
+          `Answers from ${req.params.shortname.toUpperCase()}`,
+        )
+        $('meta[property="og:description"]').attr(
+          'content',
+          `Answers from ${req.params.shortname.toUpperCase()}`,
+        )
+      }
+      res.header('content-type', 'text/html').send($.html())
+    })()
```
LoneRifle: Why wrap it up like this? You can simply make the `(req, res) => { ... }` handler async

server/src/bootstrap/index.ts
```diff
@@ -194,9 +198,100 @@ if (baseConfig.nodeEnv === Environment.Prod) {
...
+  })
+
+  app.get('*', (req, res) => {
+    const $ = cheerio.load(index)
+    $('meta[property="og:url"]').attr(
+      'content',
+      `${req.protocol}://${req.hostname}${req.originalUrl.toLowerCase()}`,
+    )
+    res.header('content-type', 'text/html').send($.html())
+  })
```
LoneRifle: I think this is slightly spurious, so maybe we can drop this =)

LoneRifle (changes requested):
Final thoughts:
- At this point, the implementation is fairly involved. Per earlier comments I made the other time, extract this out into its own module, following the structure as you find in `modules/`. So perhaps you could define something like `modules/web` and go from there with controller, route, service etc
- What do you do when the question or agency does not exist?

server/src/bootstrap/index.ts
LinHuiqing: okays

LinHuiqing:
> * What do you do when the question or agency does not exist?

For this, what should we see actually? Should it be the default or some kind of "not found" page? Currently not found pages give the default preview too :/ 

Tangent: https://staging.ask.gov.sg/agency/no gives a blank page rather than a not found page. This is a bug, right?

LoneRifle:
It's a missing feature, to be more precise. I would think that you could basically serve the not found page when you can't find the agency or question, and return http status code 404, thereby also solving #256

server/src/routes/index.ts
```diff
@@ -58,5 +61,6 @@ export const api = (options: ApiRouterOptions): express.Router => {
   router.use('/posts/answers', routeAnswers(options.answers))
   router.use('/agencies', routeAgencies({ controller: options.agency }))
   router.use('/enquiries', routeEnquiries({ controller: options.enquiries }))
+  router.use('/web', routeWeb({ controller: options.web }))
```
LoneRifle: This isn't an api route..

LoneRifle (approved):

PR #501, opened 2021-10-06
feat(agencies): add display ordering of tags

### LINKED ISSUE #490: Enforce ordering of tags
Alphabetical ordering of tags is generally incoherent, contrary to popular belief, owing to users having no intrinsic idea of possible topics. Allow agencies to specify an order of tags by id, with a view to migrating to ordering by category in due course.

LoneRifle:
## Problem

Closes #490 

## Solution

Allow agencies to specify a display order by tag id

- Define displayOrder as JSON (MySQL does not support DataTypes.ARRAY)
- Add necessary Sequelize migration
- Use displayOrder to determine order of tag, depending on if a tag
  is specified. Fallback on alphabetical ordering as needed

Take the opportunity to move to using shared types

## Screenshot
image

## Tests

- [ ] Ensure that display order for EMA in staging is not alphabetical, per screenshot

## Deploy Notes

- `20211006020849-add-agency-order` - already invoked in staging and prod

### DIFF
--- a/client/src/components/TagPanel/TagPanel.component.tsx
+++ b/client/src/components/TagPanel/TagPanel.component.tsx
@@ -14,7 +14,7 @@ import * as FullStory from '@fullstory/browser'
 import { ReactElement } from 'react'
 import { useQuery } from 'react-query'
 import { Link as RouterLink, useParams } from 'react-router-dom'
-import { TagType } from '~shared/types/base'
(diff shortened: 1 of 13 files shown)


LinHuiqing (approved):
lgtm i think

LinHuiqing:
might want to add to the test section the particular order we're looking for? at least for staging EMA

PR #502, opened 2021-10-06
fix(post): adjust colors for readability

LoneRifle:
## Problem

When clicking through to a question, answer text and related question entries are
unreadable due to bad use of colour

## Solution

- Convert AnswerItem to tsx
- Remove spurious styling in AnswerItem.styles
- Set answer text color as secondary 900, related text as primary 900

## Screenshot
image


LinHuiqing (approved):
lgtm

PR #513, opened 2021-10-07
fix(sort): fix 'Most recent' and search result sorting

LinHuiqing:
## Problems

1. 'Most recent' sorting was based on `createdAt` attribute when it should be by `updatedAt` attribute. 
2. Search results were sorted by `createdAt` attribute rather than relevance.

## Solution

1. Change sort function in post service to use `updatedAt`.
2. Remove `.sort(sortByCreatedAt)` done to search results.

## Tests

1. Update a post which was previously not at the top of the 'Most recent' sort. The post should now show up as the top result when sorting by 'Most recent'.
2. Create a post which has a prefix of a word _x_ which appears in other questions or descriptions. When _x_ is entered in to the search bar, questions with the full word should be ranked above the one with just the prefix.

### DIFF
--- a/client/src/pages/SearchResults/SearchResults.component.jsx
+++ b/client/src/pages/SearchResults/SearchResults.component.jsx
@@ -17,7 +17,6 @@ import {
   listPosts,
   LIST_POSTS_FOR_SEARCH_QUERY_KEY,
 } from '../../services/PostService'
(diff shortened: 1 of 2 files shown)


LoneRifle (approved):
lgtm, merge when ready

PR #514, opened 2021-10-07
feat(view-all): add 'View all questions' button on main (and agency) pages

### LINKED ISSUE #352: Update pagination revised design (AskGov Design Master v1)
On main (/agency) page:

- Top 10 questions by popularity
- Load more button at the end

On tag-specific page OR after clicking load more:

- pagination with 30 questions per page

LinHuiqing:
## Problem

Cluttered UI with lots of scrolling which users have to do.

Closes #352

## Solution

As according to [design on AskGov Design Master v1](https://www.figma.com/file/Y2sqYzAkssq5xSFTN9KZeX/AskGov-Design-Master-v1?node-id=5592%3A25597), following the changes were made: 

1. on main (/agency) pages: show top 10 with the 'View all questions' button at the end
2. on tag-specific page/ when 'View all questions' button is clicked: paginate with 30 questions per page

**Improvements**:

- Removed spurious QuestionsList styling sheet (it was empty...)

## Before & After Screenshots

**BEFORE**:

There were 50 questions per page. Users can browse through paginated questions with the following controls.

Pagination on mobile:

Screenshot 2021-10-07 at 4 22 31 PM

Pagination on desktop:

Screenshot 2021-10-07 at 4 22 51 PM

**AFTER**:

On main (/agency) pages, show top 10 questions. The following button is added to the end of the list for users to view more.

Screenshot 2021-10-07 at 4 25 23 PM

On tag-specific page/ when 'View all questions' button is clicked, pagination controls are the same as before.

## Tests

1. Navigate to a main (/agency) page - there should be only 10 questions with the 'View all questions' button at the end.
2. Click 'View all questions' - page should show at most 30 questions per page.
3. Click any tag - page should show paginated questions with at most 30 questions per page.
4. Try to break this by clicking around?

### DIFF
--- a/client/src/pages/HomePage/HomePage.component.jsx
+++ b/client/src/pages/HomePage/HomePage.component.jsx
@@ -244,8 +244,9 @@ const HomePage = ({ match }) => {
           ) : (
             <QuestionsListComponent
               sort={sortState.value}
-              tags={agencyAndTags}
-              pageSize={50}
+              agency={match.params.agency}
+              tags={queryState}
+              pageSize={30}
             />
           )}
         </Box>
--- a/client/src/components/QuestionsList/QuestionsList.component.tsx
+++ b/client/src/components/QuestionsList/QuestionsList.component.tsx
@@ -1,27 +1,38 @@
 import { Center } from '@chakra-ui/layout'
-import { useState } from 'react'
+import { Button, Text } from '@chakra-ui/react'
+import { useState, useEffect } from 'react'
 import { useQuery } from 'react-query'
 import { listPosts, LIST_POSTS_QUERY_KEY } from '../../services/PostService'
+import { mergeTags } from '../../util/tagsmerger'
+import Pagination from '../Pagination'
 import PostListComponent from '../PostList/PostList.component'
 import Spinner from '../Spinner/Spinner.component'
-import './QuestionsList.styles.scss'
-import Pagination from '../Pagination'
 
 interface QuestionsListProps {
   sort: string
+  agency: string
   tags: string
   pageSize: number
 }
 
 const QuestionsList = ({
   sort,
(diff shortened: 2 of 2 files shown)


client/src/pages/HomePage/HomePage.component.jsx
```diff
@@ -244,8 +244,9 @@ const HomePage = ({ match }) => {
           ) : (
             <QuestionsListComponent
               sort={sortState.value}
-              tags={agencyAndTags}
-              pageSize={50}
+              agency={match.params.agency}
+              tags={queryState}
```
LoneRifle, line 248: are you planning ahead for when agency and tags are separate entities?

LoneRifle (approved):
lgtm otherwise

client/src/pages/HomePage/HomePage.component.jsx
LinHuiqing: Not really? I separated so the QuestionList component is aware of the difference between agency and tags, because agency main page and the AskGov main page should behave in the same way.

LinHuiqing: Is there a better approach for this?

LoneRifle: I would have left it alone, usually, due to separation of concerns - the QuestionList does not have to know the distinction between the two.

Merge when ready!

LinHuiqing: Hm I think since pagination/ button at the bottom of the PostList is in QuestionsList, it should be QuestionsList's responsibility to manage the page size and displaying the button when its on the main pages? And this logic isn't related to anything else in the home page after all... 

But okay it doesn't really matter that much :joy:

PR #515, opened 2021-10-07
fix(edit-button): invalidate post query on delete

LoneRifle:
## Context

Ensure that react-query invalidates the query for the post that has been deleted


LinHuiqing (approved):
lgtm

PR #539, opened 2021-10-11
fix(text-overflow): fix text overflow issue for accordian button

LinHuiqing:
## Problem

Accordian button text still overflows. This means that topic pages where there are long topics will be ugly :(

## Solution

Remove fixed height and add 24px padding top and 16px padding bottom.

## Before & After Screenshots

**BEFORE**:

image

**AFTER**:

Screenshot 2021-10-11 at 11 48 43 AM

## Tests

Go to page where a topic is selected. Resize viewport such that text overflows to next line and check that it looks something like the screenshot (ie still has padding around text).


LoneRifle (approved):

wanlingt (approved):
lgtm

PR #557, opened 2021-10-12
fix(search): wraparound text for autosuggest entries

LoneRifle:
## Problem and Solution 
- Change fixed height and truncated text overflow to variable height,
  padding and text wraparound
- Take the chance to use chakra-specified theme colours where possible

## Before & After Screenshots

| Before | After |
|-|-|
| image | image |

### DIFF
--- a/client/src/components/SearchBox/SearchBox.styles.scss
+++ b/client/src/components/SearchBox/SearchBox.styles.scss
@@ -1,4 +1,3 @@
-@import '../../styles/colours';
 @import '../../styles/misc';
 @import '../../styles/text';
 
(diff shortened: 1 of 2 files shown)


LinHuiqing (approved):
lgtm! so the only the truncation was simply because of the height set? hahah

LoneRifle:
> lgtm! so the only the truncation was simply because of the height set? hahah

and a bunch of other CSS attributes. see changes.

LinHuiqing:
Ah right okay, just saw it. Could we remove the black outline while we're at the search box now?

client/src/components/SearchBox/SearchBox.styles.scss
```diff
@@ -17,8 +16,8 @@
       flex: 1;
 
       .search-box {
-        color: $neutral-900;
-        background-color: $white;
+        color: var(--chakra-colors-neutral-900);
+        background-color: white;
         border: 1px solid var(--chakra-colors-gray-800);
```
LinHuiqing:
```suggestion
```
should be this line?

PR #566, opened 2021-10-13
fix(post): change title colour to pri.500

LoneRifle:
Self-explanatory

| Before | After |
|-|-|
| image | image |


LinHuiqing (changes requested):
The colour of the wrong text was changed:

Screenshot 2021-10-13 at 9 54 16 AM

LinHuiqing:
> The colour of the wrong text was changed:

Should be line 17 of `AnswerItem` instead.

LinHuiqing (approved):
lgtm

PR #568, opened 2021-10-13
fix(links): fallback to https if cannot parse url

LoneRifle:
## Problem

AskGov's frontend crashes if it cannot parse a malformed URL found in the content of the post

## Solution

Since we parse the URL only to obtain the protocol, gracefully fallback on `https:` if we fail to do so

TODO:
Prevent the user from creating malformed users, and warn the public or users if bad URLs are encountered
in content

## Tests
- Create a question with a badly formed URL 
- Verify that the page continues to render


LinHuiqing (approved):
lgtm, TIL you can do that haha

PR #570, opened 2021-10-13
feat(sitemap): add sitemap

### LINKED ISSUE #400: Add sitemap file
## Problem

Search crawlers may not be able to understand the structure of the site, resulting more time spent crawling unnecessary URLs on the site. This waste of time translates to a waste of crawl budget, meaning that search crawlers might not be able to index the whole site in time.

A sitemap helps in our case as our agency pages are in silos.

## Solution

Attach the site map for search crawlers to read and understand the site.

## Additional Context

- Google's crash course to sitemaps: https://developers.google.com/search/docs/advanced/sitemaps/overview
- Possible tool: https://github.com/kuflash/react-router-sitemap

LinHuiqing:
## Problem

Search crawlers may not be able to understand the structure of the site, resulting more time spent crawling unnecessary URLs on the site. This waste of time translates to a waste of crawl budget, meaning that search crawlers might not be able to index the whole site in time.

A sitemap helps in our case as our agency pages are in silos.

Closes #400 

## Solution

Use `express-sitemap-xml` to generate and serve sitemap. Sitemap is served at `/sitemap.xml`. Sitemap checkers (e.g. https://seositecheckup.com/tools/sitemap-test) should be able to see sitemap.

**Features**:

- add all paths which members of public should be able to see to the sitemap

## Before & After Screenshots

**BEFORE**:

When we `GET` `/sitemap.xml`, users will be directed to `/not-found`.

Sitemap checkers are unable to detect the sitemap either:

Screenshot 2021-10-13 at 1 58 10 PM

**AFTER**:

Sitemap file is served at `/sitemap.xml`.

Screenshot 2021-10-13 at 1 55 05 PM

Sitemap checker is able to detect the sitemap.

Screenshot 2021-10-13 at 1 54 34 PM

## Tests

1. Check `/sitemap.xml` to ensure that a sitemap is served and that it includes static paths (visible to public), post pages and agency pages.
2. Use a sitemap checker to check that sitemap can be detected.

### DIFF
--- a/server/src/bootstrap/index.ts
+++ b/server/src/bootstrap/index.ts
@@ -59,6 +59,7 @@ import {
   User,
 } from './sequelize'
 import sessionMiddleware from './session'
+import { SortType } from '../types/sort-type'
 
 export { sequelize } from './sequelize'
 export const app = express()
@@ -195,7 +196,7 @@ if (baseConfig.nodeEnv === Environment.Prod) {
 
 app.use('/api/v1', api(apiOptions))
 
-const getSitemapUrls = () => {
+const getSitemapUrls = async () => {
   const visibleStaticPaths = ['/', '/terms', '/privacy']
   const sitemapLeaves: SitemapLeaf[] = []
   for (const path of visibleStaticPaths) {
@@ -204,7 +205,16 @@ const getSitemapUrls = () => {
       lastMod: true,
     })
   }
-
+  const { posts: allPosts } = await postService.listPosts({
+    sort: SortType.Top,
+    tags: '',
+  })
+  for (const post of allPosts) {
+    sitemapLeaves.push({
+      url: `/questions/${post.id}`,
+      lastMod: true,
+    })
+  }
   return sitemapLeaves
 }
 
--- a/server/src/bootstrap/index.ts
+++ b/server/src/bootstrap/index.ts
@@ -33,6 +33,7 @@ import { routeWeb } from '../modules/web/web.routes'
 import { WebService } from '../modules/web/web.service'
 import { api } from '../routes'
 import { RecaptchaService } from '../services/recaptcha/recaptcha.service'
+import { SortType } from '../types/sort-type'
 import { bannerConfig } from './config/banner'
 import { baseConfig, Environment } from './config/base'
 import { datadogConfig } from './config/datadog'
@@ -59,7 +60,6 @@ import {
   User,
 } from './sequelize'
 import sessionMiddleware from './session'
-import { SortType } from '../types/sort-type'
 
 export { sequelize } from './sequelize'
 export const app = express()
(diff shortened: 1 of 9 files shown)


server/package.json
```diff
@@ -32,6 +32,7 @@
     "datadog-winston": "^1.5.1",
     "express": "^4.17.1",
     "express-session": "^1.17.2",
+    "express-sitemap-xml": "^3.0.0",
```
LoneRifle: File the dep change as a separate PR so that you don't have to have a headache with merge conflicts

server/src/bootstrap/index.ts
```diff
@@ -194,12 +196,45 @@ if (baseConfig.nodeEnv === Environment.Prod) {
 
 app.use('/api/v1', api(apiOptions))
 
+const getSitemapUrls = async () => {
```
LoneRifle: I would implement this into its own module..

LoneRifle (changes requested):
Don't forget your tests =)

LoneRifle:
One other thing - `PostService.listPosts()` will be relatively expensive to invoke. Can you figure a way to minimise the number of invocations? If you are going to approach this with caching, you may need to devise a strategy for cache invalidation , and if so, you need to ensure that the db queries (if any!) you make to ensure sitemap freshness is cheap to run.

server/package.json
LinHuiqing: Okays, filed PR #571.

LinHuiqing:
> One other thing - `PostService.listPosts()` will be relatively expensive to invoke. Can you figure a way to minimise the number of invocations? If you are going to approach this with caching, you may need to devise a strategy for cache invalidation , and if so, you need to ensure that the db queries (if any!) you make to ensure sitemap freshness is cheap to run.

Thanks for bringing this up, I thought about these considerations but eventually left it as the current implementation due to the following reasons: 

1. Based on the [`express-sitemap-xml`](https://github.com/feross/express-sitemap-xml) documentation: 

    > The `getUrls` function is called at most once per 24 hours. The resulting sitemap(s) are cached to make repeated HTTP requests faster.
    
    This means that though the `PostService.listPosts()` function is expensive, it will only be called once per day when the server is up and running.

2. Google's search crawlers only use sitemaps as a recommendation according to [Google Search Central's documentation](https://developers.google.com/search/docs/advanced/sitemaps/overview):

    > Using a sitemap doesn't guarantee that all the items in your sitemap will be crawled and indexed, as Google processes rely on complex algorithms to schedule crawling. However, in most cases, your site will benefit from having a sitemap, and you'll never be penalized for having one.

    So to my knowledge, it would not be that detrimental if our site is slightly outdated as Google will still be able to crawl the new/ updated pages as long as they are linked somewhere in our site.

Considering the above, do you think we should still try to minimise the number of invocations and ensure sitemap freshness?...

server/src/bootstrap/index.ts
LinHuiqing: okay done 👍

LoneRifle:
Good points! I would imagine that there are other crawlers out there, but that should be mostly handled by the package, as you've pointed out.

LoneRifle (approved):
lgtm. Do provide unit tests for the module if you can spare the time.

Issue #225, opened 2021-09-06
Convert usage of Model and ModelCtor to ModelDefined


mantariksh:
Advantages of `ModelDefined`:
- Allows you to define interfaces which specify the object shape without extending from `Model`
- Allows you to specify which attributes the model will have once it is created, and which attributes must be specified at the point of creation (e.g. if a model has a default value for a particular attribute, it will have that value once it is created but it doesn't need to be specified at the point of creation)

LoneRifle:
The current definitions in Sequelize don't recognise that a `Model<M>` would also have object properties of M in addition to Model. A custom type will have to be introduced that effectively does this:
```ts
type ModelDef<M, C = M> = ModelCtor<Model<M, C> & M>
```

PR #581, opened 2021-10-14
refactor(posts): move to ModelDef

LoneRifle:
## Problem

Post currently has an object definition that extends Sequelize.Model, in addition
to a shared Post type defined earlier. The codebase should move to use the latter.

Part of #225 

## Solution

Replace use of ModelCtor with ModelDef, carefully silencing tsc errors
arising from the definition of PostTag without any attributes. We can
safely do this since we will be defining the foreign key relations for
the model later on.


wanlingt (approved):
lgtm

PR #604, opened 2021-10-18
feat(topics): add Topics api endpoints

### LINKED ISSUE #398: Add API endpoints for Topics
## Problem
No API endpoints exist for topics, refer to [design doc](https://docs.google.com/document/d/1JFmO4rttnfeGFl15gVh63SVRk0w_jp2d6r4B7GVjssI/edit?usp=sharing) for more details

wanlingt:
## Problem

Topics do not have API endpoints

Closes #398 

## Solution

Add the following API endpoints:
```
GET /:agencyId
POST /
DELETE /:id
PUT /:id
```
These allow us to
- retrieve nested list of topics belonging to an agency
- create / update / delete a topic - for users belonging to the same agency as the topic

Additionally, `verifyUserCanModifyTopic` and `verifyUserInAgency` were added to AuthService.

## Tests

Run the unit tests for topics and agency


server/src/modules/topics/topics.routes.ts
```diff
@@ -0,0 +1,80 @@
...
+  const router = express.Router()
+
+  /**
+   * Lists all topics corresponding to an agency
+   * @route   GET /api/topics/agency/:agencyId
+   * @returns 200 with list of nested topics
+   * @returns 400 with database error
+   */
+  router.get(
+    '/:agencyId',
```
LoneRifle:
```suggestion
    '/agency/:agencyId',
```
To line up with the jsdoc

LoneRifle: A clearer alternative may be to have a path like `/agencies/:agencyId/topics`, but that would require you to modify the agency route, injecting it with the topics controller

LoneRifle (approved):
lgtm otherwise

server/src/modules/topics/topics.routes.ts
wanlingt: went with the latter suggestion - changed the route to `/agencies/:agencyId/topics`

PR #605, opened 2021-10-18
fix(web): sanitize html of question page description content

LinHuiqing:
## Problem

I forgot to sanitize the description after I refactored the code into the web module...

## Solution

- use `sanitize-html` to sanitize first answer of a post before putting it into the description tags
- refactor code so index file is only loaded once

## Before & After Screenshots

**BEFORE**:

On [Facebook Sharing Debugger](https://developers.facebook.com/tools/debug/), raw description fields appear as white space:

Screenshot 2021-10-18 at 6 12 55 PM

**AFTER**:

On [Facebook Sharing Debugger](https://developers.facebook.com/tools/debug/), raw description fields are populated:

Screenshot 2021-10-18 at 6 24 39 PM

## Tests

Check a question page on Facebook Sharing Debugger. The raw tags should have text for the description fields.

### DIFF
--- a/server/src/modules/web/web.controller.ts
+++ b/server/src/modules/web/web.controller.ts
@@ -56,6 +56,7 @@ export class WebController {
       })
       if (agency.isOk()) {
         const agencyPage = await this.webService.getAgencyPage(
+          this.index,
           req.params.shortname,
           agency.value.longname,
         )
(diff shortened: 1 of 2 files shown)


LoneRifle (approved):
Lgtm. Long term, we might want to consider being ok with questions without answers, since there may be use cases for that for members of the public.

PR #612, opened 2021-10-19
fix(sort-icon): change sort icon to BiSortAlt2

LinHuiqing:
Change sort icon to BiSortAlt2.

## Before & After Screenshots

**BEFORE**:

Screenshot 2021-10-19 at 9 56 59 AM

**AFTER**:

Screenshot 2021-10-19 at 9 56 38 AM

## Tests

Go to a topics page and check the icon on the right of the sorting field.


wanlingt (approved):
lgtm

PR #613, opened 2021-10-19
refactor(post): add agencyId

### LINKED ISSUE #3: Authorization policy to manage content inconsistent
There are currently two prevailing policies used to manage content:

- Answers can be edited by users whose agency tags intersect with those found on the post
- Posts can _only_ be edited by users _with the same exact set of tags_ as the post's tags

One of these policies should be chosen, and implemented throughout the codebase

### LINKED ISSUE #603: Add foreign key from Agency to Post
As part of efforts to migrate from a tag-centric to a topic-centric data model,
We should refrain from using tags to describe which Agency a Post belongs to.
Adding an `agencyId` would allow us to enforce this relationship better, and allow
tags to be exclusively for what it is meant for - ie, marking posts by keyword.

Given that Posts will now require an agencyId to be supplied at creation, we will use
the one tied to the user. While we lose the ability to associate a Post with multiple
agencies, this feature has never actually been used in production, and can be dropped.

Implementing this issue will involve the following:
- Changing the data model to include the proposed foreign key constraint
- Ensuring that newly created posts will contain the user's agencyId
- Changing retrieval of posts by agency by using agencyId

This issue may also incorporate #3 .

LoneRifle:
## Problem

Closes #3 
Closes #603

## Solution

Establish a relationship between Post and Agency to facilitate work
to migrate to topics

- Add agencyId to shared and sequelize type
- Infer agencyId from user when creating posts
- Drop isArray validator for Agency.displayOrder, invalid
- Match permissions for users to manage posts on agencyIds
- Provide test coverage for posts
  - Take the trouble to introduce dep injection to UserService to provide more realistic tests

## Tests

- [ ] verify users can continue to create/update/delete posts

## Deploy Notes

**New scripts**:
- `20211019053050-create-post-agency-id` : invoke this migration first, then perform backfill to ensure agency ids for posts in place
- `20211019053110-update-post-agency-id-non-nullable`: enforce Posts to have agency ids

### DIFF
--- a/server/migrations/20211019053050-create-post-agency-id.js
+++ b/server/migrations/20211019053050-create-post-agency-id.js
@@ -1,0 +1,20 @@
+'use strict'
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    queryInterface.addColumn('posts', 'agencyId', {
+      allowNull: true,
+      type: Sequelize.INTEGER,
+      references: {
+        model: 'agencies',
+        key: 'id',
+      },
+      onUpdate: 'CASCADE',
+      onDelete: 'CASCADE',
+    })
+  },
+
+  down: async (queryInterface) => {
+    await queryInterface.removeColumn('posts', 'agencyId')
+  },
+}
--- a/server/migrations/20211019053110-update-post-agency-id-non-nullable.js
+++ b/server/migrations/20211019053110-update-post-agency-id-non-nullable.js
@@ -1,0 +1,17 @@
+'use strict'
+
+module.exports = {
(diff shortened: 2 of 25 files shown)


wanlingt (approved):
lgtm! 

I assume we'll remove the agency tag linked to each post in a separate PR where we remove the idea of agency tags totally?

LoneRifle:
> lgtm!
> 
> I assume we'll remove the agency tag linked to each post in a separate PR where we remove the idea of agency tags totally?

Yup; a surgical removal of the agency tag rows followed by removing the corresponding TagType value from the codebase

PR #623, opened 2021-10-21
chore(post): add scripts to remove and readd agency id fk

LoneRifle:
## Problem and Solution

MySQL does not take kindly to column changes when a foreign key is
present, so provide scripts that would cleanly remove and readd the
fk before/after the setting of agency id to not null

## Deploy notes

Already invoked in production, no further action needed


wanlingt (approved):
lgtm

PR #625, opened 2021-10-21
refactor(post): retrieve answerable posts by agencyId

LoneRifle:
## Problem
AskGov is still using agency tags for client-side permissions checks and
retrieving answerable posts

## Solution
Stop using agency tags to do client-side permissions checks and for
listing answerable posts, favouring the use of agency id instead

- Add agency to LoadUserDto, agencyId to responses from PostService
- Rework `PostService.listAnswerablePosts` to use `agencyId`
- Make use of `user.agency` or `user.agencyId` throughout client,
  dropping the need for complicated tag comparisons

Retrieving posts by agency id will be undertaken as part of a wider PR for topics - #630

### DIFF
--- a/client/src/api/types/post.ts
+++ b/client/src/api/types/post.ts
@@ -4,6 +4,7 @@ import { User as UserBaseDto, PostStatus } from '~shared/types/base'
 
 export type BasePostDto = BaseModelParams & {
   userId: number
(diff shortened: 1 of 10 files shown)


wanlingt (approved):
lgtm

PR #630, opened 2021-10-22
feat(post): modify endpoints to include Topics

### LINKED ISSUE #399: Modify Posts API endpoints to include topics
## Problem
Endpoints for Posts API do not account for the inclusion of topics.

## Solution
Refer to [design doc](https://docs.google.com/document/d/1JFmO4rttnfeGFl15gVh63SVRk0w_jp2d6r4B7GVjssI/edit#heading=h.oayaz2jd72g) for more details

wanlingt:
## Problem

Topics are not included in Post endpoints

Closes #399 

## Solution
Modify the following:
- `listPosts`
  - Add `topics` as a param to filter posts. More than 1 topic may be specified. Returns posts belonging to the specified topic(s) and their child topics.
  - Add `agencyId` as a param to filter posts. Validity checks for agency are already done on the client-side.
  - Remove merging of agency and tags; `agency` is no longer included in `tags` param
  - Refactor queries - only one query is used to retrieve posts. When more than 1 topic / tag is specified, posts belonging to those topics / tags are returned, instead of posts which match that exact combination of topics / tags
- `listAnswerablePosts`
  -  Refactor queries in the same manner as `listPosts`
  - Refactor splitting of tags, move splitting of tags from `post.routes` to `post.service`
- `getSinglePost`
  - Add `topic` as a param 
- `createPost`
  - Add `topicname` as a param 
  - Error thrown if no topics and tags are specified; at least one topic or tag is necessary to create a post
- `updatePost`
  - Add `topicname` as a param 
- `getRelatedPosts`
  - Look for posts belonging to the same topic, followed by posts with the most number of overlapping tags, followed by number of views

**Improvements**:

- Replace redundant `PostWithRelations` with `Posts`

### For `getRelatedPosts`

**BEFORE**:
image 
**AFTER**:
Screenshot 2021-10-22 at 5 45 13 PM

## Tests
- Adding and removing tags from a post should work the same as before
- If multiple tags are specified when filtering posts by tags, posts returned should belong to at least one of the specified tags. Previously, posts returned had to have exact same combination of tags.


client/src/api/types/post.ts
```diff
@@ -24,6 +25,7 @@ export type GetPostsDto = {
 
 export type CreatePostReqDto = Pick<BasePostDto, 'title' | 'description'> & {
   tagname: string[]
+  topicname: string
```
LoneRifle:
```suggestion
  topicname: string
```
- Given that topics are optional, we should allow for null topics to be specified
- I've sorely regretted being unable to intervene with casing of parameters back when others were handling the codebase. Could we use camelCasing here please?

client/src/components/QuestionsList/QuestionsList.component.tsx
```diff
@@ -7,15 +7,15 @@ import {
...
 } from '../../services/PostService'
-import { mergeTags } from '../../util/tagsmerger'
 import Pagination from '../Pagination'
 import PostListComponent from '../PostList/PostList.component'
 import Spinner from '../Spinner/Spinner.component'
 
 interface QuestionsListProps {
   sort: string
-  agency: string
+  agency: number
```
LoneRifle: For clarity..
```suggestion
  agencyId: number
```

client/src/components/QuestionsList/QuestionsList.component.tsx
```diff
@@ -25,35 +25,36 @@ const QuestionsList = ({
...
+            topics,
             page,
             size: pageSize,
           }),
       }
     : {
         queryKey: [
           LIST_POSTS_QUERY_KEY,
-          { sort, agencyAndTags, page, pageSize },
+          { sort, agency, tags, topics, page, pageSize },
```
LoneRifle: Similarly..
```suggestion
          { sort, agencyId, tags, topics, page, pageSize },
```

client/src/components/QuestionsList/QuestionsList.component.tsx
```diff
@@ -25,35 +25,36 @@ const QuestionsList = ({
...
         ],
-        queryFn: () => listPosts(sort, agencyAndTags, page, pageSize),
+        queryFn: () => listPosts(sort, agency, tags, topics, page, pageSize),
```
LoneRifle: and also..
```suggestion
        queryFn: () => listPosts(sort, agencyId, tags, topics, page, pageSize),
```

client/src/pages/EditForm/EditForm.component.tsx
```diff
@@ -69,6 +69,7 @@ const EditForm = ({ match }: EditFormProps): JSX.Element => {
       description: data.postData.description,
       title: data.postData.title,
       tagname: data.tags,
+      topicname: '', //TODO: update when topics selection is introduced in frontend
```
LoneRifle: If the client type supports using null, you should likewise use null
```suggestion
      topicName: null, //TODO: update when topics selection is introduced in frontend
```

server/src/modules/post/post.service.ts
```diff
@@ -168,127 +235,190 @@ export class PostService {
...
   listPosts = async ({
     sort,
+    agency,
     tags,
+    topics,
     page,
     size,
   }: {
     sort: SortType
+    agency: number
```
```suggestion
    agencyId: number
```
and elsewhere

server/src/modules/post/post.routes.ts
```diff
@@ -37,30 +37,12 @@ export const routePosts = ({
...
-        2. repeated query param eg. tags=ema&tags=ogp
-        
-        eg. if query params are tags=parking,parking-activate&tags=ogp
-        req.query.tags = ['parking, parking-activate', 'ogp'] as body-parser accept query param format 2
-        this custom santiser transforms req.query.tags into ['parking', 'parking-activate', 'ogp']
-        */
-        const tagsArray: string[] = [].concat(value)
-        const nestedTags = tagsArray.map((tags) => tags.split(','))
-        return ([] as string[]).concat(...nestedTags)
-      }),
```
LoneRifle, line 57: wait, why was this removed? It's still needed to validate and sanitise tags. If anything you would need to implement similar for topics too.

server/src/modules/post/post.service.ts
```diff
@@ -84,50 +96,105 @@ export class PostService {
...
+              where: {
+                id: tags,
+              },
+            },
+            {
+              model: this.Topic,
+              attributes: [],
+              required: false,
+              where: { id: post.topicId },
+            },
```
LoneRifle: You can avoid the join simply by matching on `Post.topicId` in `where`

server/src/modules/post/post.service.ts
```diff
@@ -84,50 +96,105 @@ export class PostService {
...
+    const relatedPosts = topic
+      ? await this.Post.findAll({
+          attributes: {
+            include: [
+              [
+                Sequelize.fn('COUNT', Sequelize.col('topicId')),
+                'relatedTopics',
+              ],
+              [Sequelize.fn('COUNT', Sequelize.col('tags.id')), 'relatedTags'],
+            ],
```
LoneRifle: This would look a little complicated, but to avoid the double `Post.findAll` call, you could specify the desired attributes and order in a separate variable, then spread into the Post, ie,
```ts
const findOptions = post.topicId
  ? { attributes: { include: [...] }, order: [...] }
  : ..... 
```

Alternatively, specify the original findOptions as if there is no topic, and if there is indeed a topic, change `attributes.include` and `order` accordingly.

Then spread into `Post.findAll`:
```ts
const relatedPosts = await this.Post.findAll({
  // ...
  ...findOptions,
  // ......
})
```

server/src/modules/post/post.service.ts
```diff
@@ -168,127 +235,190 @@ export class PostService {
...
+      while (currGenChildTopics.length) {
+        const currGenChildTopicIds = currGenChildTopics.map((topic) => topic.id)
+        const nextGenChildTopics = agencyTopics.filter(
+          (topic) =>
+            !!topic.parentId && currGenChildTopicIds.includes(topic.parentId),
+        )
+        parentAndChildTopics = parentAndChildTopics.concat(nextGenChildTopics)
+        currGenChildTopics = nextGenChildTopics
+      }
+    }
```
LoneRifle, line 271: How does this return a tree?

server/src/modules/post/post.service.ts
```diff
@@ -168,127 +235,190 @@ export class PostService {
...
     totalItems: number
   }> => {
-    // basic
-    let tags_unchecked: string[] = []
-
+    // split tags
+    let tagsUnchecked: string[] = []
     if (tags) {
-      tags_unchecked = tags.split(',')
+      tagsUnchecked = tags.split(',')
```
LoneRifle: Generally, the service should not be left to handle the parsing of `tags` or `topics` (and in fact, this should already be done by the validation logic in post.routes)

LoneRifle (changes requested):
quite a bit to get through, so take your time. Ask me to come along to pair if you need.

server/src/modules/post/post.routes.ts
wanlingt: The splitting of tags/topics is now done in post.service - previously there was an inconsistency (for listPosts it was done in post.service, but for listAnswerablePosts it was done in post.routes, so I moved it). Is it better if the splitting of tags/topics is done in post.routes though?

wanlingt: ok ignore my reply! just saw your comment below - will move to post.routes

server/src/modules/post/post.service.ts
wanlingt: it doesn't, I didn't think a tree structure was necessary as I just wanted a list of relevant topicIds for the where query in listPosts

server/src/modules/post/post.routes.ts
```diff
@@ -20,7 +20,48 @@ export const routePosts = ({
...
+      
+      eg. if query params are topics=parking,parking-activate&topics=ogp
+      req.query.topics = ['parking, parking-activate', 'ogp'] as body-parser accept query param format 2
+      this custom santiser transforms req.query.topics into ['parking', 'parking-activate', 'ogp']
+      */
+        const topicsArray: string[] = [].concat(value)
+        const nestedTopics = topicsArray.map((topics) => topics.split(','))
+        return ([] as string[]).concat(...nestedTopics)
+      }),
+    ],
```
LoneRifle: apply DRY here and have the sanitisers held in a separate array, `sanitizeTagsAndTopics` or similar, then refer to them for both the answerable and public posts endpoints

LoneRifle (approved):
lgtm, good work!

PR #645, opened 2021-10-25
refactor(palette): update neutral, success, warning to v2

LoneRifle:
## Problem and Solution

As part of wider work to move to the v2 design system, we have to change the colour palettes.

This PR updates colour palettes for neutral, success and warning colours to line up with v2.
Given the relative minor changes in colours, it makes sense to incorporate them to the codebase
immediately. It is hence being submitted ahead of changes to the secondary and primary palettes,
which may incur more changes.

Take the chance to remove references to SCSS colours, instead using Chakra vars for everything


LinHuiqing (approved):
lgtm, I'm assuming @khaleedah took a look at it while it was on staging?

PR #651, opened 2021-10-25
build(node): use v16

LoneRifle:
## Problem and Description

Use the latest LTS node as part of a 
wider effort to do so across @opengovsg


LinHuiqing (approved):
lgtm

PR #672, opened 2021-10-27
feat(tags): Add cards view for topics selection (desktop and tablet)

### LINKED ISSUE #550: Change topics selection to cards view on landing page for tablet and desktop
Refer to [Figma](https://www.figma.com/file/Y2sqYzAkssq5xSFTN9KZeX/AskGov-Design-Master-v1?node-id=5752%3A7258)

wanlingt:
## Problem

The views for topic selection are the same across all devices. On larger screens, this results in wasted space.

Closes #550 

## Solution

For tablets and desktops, use grids to display topics in individual boxes ([figma reference](https://www.figma.com/file/O6lxjjfn04Ow6nn2TMStpk/AskGov-Design-Master-v2?node-id=6003%3A70135))

## Before & After Screenshots

**BEFORE**:

https://user-images.githubusercontent.com/56983748/138998969-42407adc-e5f4-4bd0-8a63-25bf369f9725.mov

**AFTER**:

https://user-images.githubusercontent.com/56983748/138998873-c52d3903-1faa-44bb-a06c-3fb37302acf5.mov

## Tests

- Click around and check that navigation to topics still works

### DIFF
--- a/client/src/components/TagMenu/TagMenu.component.tsx
+++ b/client/src/components/TagMenu/TagMenu.component.tsx
@@ -11,6 +11,8 @@ import {
   Flex,
   Spacer,
   Stack,
+  SimpleGrid,
+  Box,
 } from '@chakra-ui/react'
 import * as FullStory from '@fullstory/browser'
 import { BiRightArrowAlt } from 'react-icons/bi'
@@ -77,7 +79,6 @@ const TagMenu = (): ReactElement => {
       )
     : useQuery(FETCH_TAGS_QUERY_KEY, () => fetchTags())
 
-  // TODO - create an AccordionItem for agency tags
   return (
     <Accordion allowMultiple allowToggle>
       <AccordionItem border="none">
@@ -125,7 +126,12 @@ const TagMenu = (): ReactElement => {
             </Flex>
           </AccordionButton>
         </h2>
-        <AccordionPanel p={0} shadow="md">
+        {/* Accordion for mobile */}
+        <AccordionPanel
+          p={0}
+          shadow="md"
+          display={{ base: 'block', sm: 'none' }}
+        >
           {isLoading && <Spinner />}
           {tags && (
             <VStack align="left" spacing={0}>
@@ -143,7 +149,6 @@ const TagMenu = (): ReactElement => {
                       w="100%"
                       textAlign="left"
                       textStyle="h4"
-                      borderBottomWidth="1px"
                       role="group"
                       _hover={{ bg: 'primary.100' }}
                       _focus={{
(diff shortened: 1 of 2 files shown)


LoneRifle:
Have you considered the use of `useBreakpointValue()`[^1] to determine which grid to use, rather than multiple AccordionPanel definitions?

[^1]: https://chakra-ui.com/docs/hooks/use-breakpoint-value

wanlingt:
> Have you considered the use of `useBreakpointValue()`[1](#user-content-fn-1-6bf84e950f99ea5ca590b82bb06d9ceb) to determine which grid to use, rather than multiple AccordionPanel definitions?
> 
> ## Footnotes
> 1. https://chakra-ui.com/docs/hooks/use-breakpoint-value [↩](#user-content-fnref-1-6bf84e950f99ea5ca590b82bb06d9ceb)

just made some changes, something like that?

client/src/components/TagMenu/TagMenu.component.tsx
```diff
@@ -77,7 +80,114 @@ const TagMenu = (): ReactElement => {
...
     : useQuery(FETCH_TAGS_QUERY_KEY, () => fetchTags())
 
-  // TODO - create an AccordionItem for agency tags
+  const isShowMobileVariant = useBreakpointValue({
+    base: true,
+    xs: true,
+    sm: false,
+    md: false,
+    lg: false,
+    xl: false,
```
LoneRifle:
```suggestion
    base: true,
    sm: false,
```

client/src/components/TagMenu/TagMenu.component.tsx
```diff
@@ -77,7 +80,114 @@ const TagMenu = (): ReactElement => {
...
+  })
+
+  const TagMenuMobile = tags && (
+    <VStack align="left" spacing={0}>
+      {tags
+        .filter(
+          ({ tagType, tagname }) =>
+            tagType === TagType.Topic && tagname !== queryState,
+        )
+        .sort((a, b) => (a.tagname > b.tagname ? 1 : -1))
```
LoneRifle: extract this to a `tagsToShow` const. In order to guarantee that `tagsToShow` is always an array of some form, consider something like...
```ts
const tagsToShow = (tags || []) // tags is possibly undefined, but if so, we default to an empty array
  .filter(...)
  .sort(...)
```

This then allows you to directly use `tagsToShow` without having to do a null check ie (`tags && `).

This also means that we will still render the VStack/SimpleGrid even if there are no tags to show, but I would regard this as a good thing since there are fewer changes to the DOM that are dependent on the application state

client/src/components/TagMenu/TagMenu.component.tsx
```diff
@@ -77,7 +80,114 @@ const TagMenu = (): ReactElement => {
...
+        .map((tag) => {
+          const { tagType, tagname } = tag
```
LoneRifle:
```suggestion
        .map(({ id, tagType, tagname }) => {
```
then replace all references to tag with the appropriate property

LoneRifle (changes requested):
lgtm otherwise. Consider my points concerning the small rewrite and if you get stuck, feel free to talk to me

LoneRifle (approved):
Merge after reviewing b6ed94c

PR #682, opened 2021-10-28
build(deps): remove framer-motion, not used

LoneRifle:
## Problem and Solution

Remove framer-motion on realising that nothing in the codebase references it


LinHuiqing (approved):
Yup, looks unused

PR #703, opened 2021-11-01
docs(readme): update to reflect shared types

### LINKED ISSUE #220: npm run server requires docker
**Describe the bug**
`npm run server` no longer works because it depends on docker, specifically the `db`. So to develop the backend, `npm run dev` is still needed. This makes the `readme` misleading as new users will think `npm run server` is all they need to work on the backend.

mdonoughe:
## Problem

_What problem are you trying to solve? What issue does this close?_

New developers may not be successful setting up their development environment.

Closes #220 

## Solution

_How did you solve the problem?_

This PR is just some documentation updates for creating a development environment.

**Improvements**:

- Specified that Docker is required for `npm run dev`
- Documented a potential issue that may be encountered when using Node.js 17

**Bug Fixes**:

- Fixed spelling of `audit-dep`
- Added missing `npm run build-frontend` in `npm run dev`. Otherwise, the server immediately exits with an error about the client index.html not existing.
- Added missing `npm run build` before `npm run server`. This is for the same reason as adding `npm run build-frontend`. It unnecessarily builds the backend, but it's simpler for somebody reading the readme.

## Tests

_What tests should be run to confirm functionality?_

Running through the instructions to set up a development environment should go more smoothly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ Optionally [DBeaver](https://dbeaver.io/download/) to view database with GUI
 
 ## Running in Development
 
-* Start running frontend, backend, maildev, localstack and mysql simultaneously
+* Start running frontend, backend, maildev, localstack and mysql simultaneously (requires Docker)
 
   ```
   npm run dev
--- a/README.md
+++ b/README.md
@@ -90,10 +90,10 @@ Optionally [DBeaver](https://dbeaver.io/download/) to view database with GUI
   docker-compose up
 
   # for backend server only
-  npm run server 
+  npm run build && npm run server
 
   # for frontend server only
-  npm run client 
+  npm run client
   ```
   
   Frontend server accessible on `localhost:3000`
--- a/README.md
+++ b/README.md
@@ -138,6 +138,12 @@ Optionally [DBeaver](https://dbeaver.io/download/) to view database with GUI
 
   If that doesn't work, try it without @'localhost' part.
 
+### Node.js Related
+
+- `Error: error:0308010C:digital envelope routines::unsupported`
+
+  Try using Node.js 16.
+
 ## API Endpoints
 
 #### Base Url - `http://localhost:5000/api/v1`
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ Optionally [DBeaver](https://dbeaver.io/download/) to view database with GUI
   ```
   npm install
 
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -11,7 +11,7 @@
     "build": "npm run build-shared && npm run build-frontend && npm run build-backend",
     "server": "npm run dev --prefix server",
     "client": "npm start --prefix client",
-    "dev": "concurrently \"./wait-for-it.sh localhost:3306 -t 0 -- npm run build-shared && sleep 5 && npm run server\" \"npm run client\" \"docker-compose up \"",
+    "dev": "concurrently \"./wait-for-it.sh localhost:3306 -t 0 -- npm run build-shared && npm run build-frontend && sleep 5 && npm run server\" \"npm run client\" \"docker-compose up \"",
```
LoneRifle:
```suggestion
    "dev": "concurrently \"./wait-for-it.sh localhost:3306 -t 0 -- npm run build-shared && sleep 5 && npm run server\" \"npm run client\" \"docker-compose up \"",
```
Building the frontend isn't actually needed, since that's covered by `npm run client`

LoneRifle (changes requested):
lgtm otherwise. Thanks for the PR!

README.md
```diff
@@ -90,10 +90,10 @@ Optionally [DBeaver](https://dbeaver.io/download/) to view database with GUI
   docker-compose up
 
   # for backend server only
-  npm run server 
+  npm run build && npm run server
```
LoneRifle:
```suggestion
  npm run build-shared && npm run server
```

package.json
mdonoughe: I thought it would be too.

`npm run client` does build the client, but it doesn't write the files to `client/build`, and even though the frontend hosted by the `npm run client` server will be used, the backend will fail to start if `client/build/index.html` doesn't exist. If you delete that directory and then do `npm run client` in one terminal, wait for it to finish, and then do `npm run server` in another terminal, you will get an error about not being able to find the client files.

Alternatively, if `server/src/bootstrap/index.ts` somehow knew it was running in a development setup it could skip trying to read `index.html`.

LinHuiqing: Thanks for bringing this up - I'll be doing the latter soon 👍

PR #704, opened 2021-11-01
refactor(answers): simplify AnswerService

### LINKED ISSUE #646: Refactor AnswersService to use dependency injection

AnswersService currently does not use dependency injection and instead imports models directly from `../../bootstrap/sequelize`. This will be troublesome when unit tests are being written.

Refactor code such that these imports:

```ts
import {
  User as UserModel,
  Post as PostModel,
  Agency as AgencyModel,
  Answer as AnswerModel,
} from '../../bootstrap/sequelize'
```

will be passed in as arguments when constructing the module instead.

LoneRifle:
## Problem

Closes #646 

## Solution

- Drop query joins no longer used in `AnswerService.listAnswers()`;
  frontend no longer queries for user/agency information for each
  answer as we've changed approach from submitted question ->
  multiple answers to agency-created question and answer

- Rework to use dependency injection and shared types

## Tests

- Ensure that answers continue to be fetched normally

### DIFF
--- a/server/src/modules/answers/answers.controller.ts
+++ b/server/src/modules/answers/answers.controller.ts
@@ -5,6 +5,7 @@ import { createLogger } from '../../bootstrap/logging'
 import { StatusCodes } from 'http-status-codes'
 import { ControllerHandler } from '../../types/response-handler'
 import { Message } from '../../types/message-type'
+import { Answer } from '~shared/types/base'
 
 const logger = createLogger(module)
 
@@ -29,16 +30,10 @@ export class AnswersController {
    * @returns 200 with array of answers
    * @returns 500 if database error occurs
    */
-  listAnswers: ControllerHandler<
-    { id: string },
-    | {
-        body: string
-        username: string
-        userId: number
-        agencyLogo: string
-      }[]
-    | Message
-  > = async (req, res) => {
+  listAnswers: ControllerHandler<{ id: string }, Answer[] | Message> = async (
+    req,
+    res,
+  ) => {
     try {
       const answers = await this.answersService.listAnswers(
         Number(req.params.id),
--- a/server/src/modules/answers/answers.service.ts
+++ b/server/src/modules/answers/answers.service.ts
@@ -1,80 +1,32 @@
-import {
-  User as UserModel,
-  Post as PostModel,
-  Agency as AgencyModel,
-  Answer as AnswerModel,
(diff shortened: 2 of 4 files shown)


server/src/modules/answers/answers.controller.ts
```diff
@@ -29,16 +30,10 @@ export class AnswersController {
...
-    { id: string },
-    | {
-        body: string
-        username: string
-        userId: number
-        agencyLogo: string
-      }[]
-    | Message
-  > = async (req, res) => {
+  listAnswers: ControllerHandler<{ id: string }, Answer[] | Message> = async (
```
LinHuiqing, line 33: If we are going for agency-centric question and answer, is there still a need for value returned to be an array?

server/src/modules/answers/answers.service.ts
```diff
@@ -1,80 +1,32 @@
...
-        } = answer
-        return {
-          ...answerWithoutUser,
-          username: displayname,
-          userId: id,
-          agencyLogo: logo,
-        }
-      })
-    } else return []
+  listAnswers = async (postId: number): Promise<Answer[]> => {
```
LinHuiqing, line 25: Same as above, should this still be returning an array instead of just an Answer object?

LoneRifle: we'll discuss this on the other comment

server/src/modules/answers/answers.controller.ts
LoneRifle: I'm not entirely sure. We _might_ want to return multiple answers for other reasons (perhaps if we want to return answers and updates to such answers over time?). The changes needed to frontend are a little invasive anyway, so perhaps we can defer the changes to response shape to a later PR.

wanlingt (approved):
lgtm

server/src/modules/answers/answers.controller.ts
LinHuiqing: Hm okay maybe this in its current form for this PR is fine. There doesn't seem to be a way for agency to input more than one answer, so it seems a little inconsistent in the codebase. Maybe in the future this will be useful then 👍

LinHuiqing: fwiw, I think the answers and updates over time might be confusing for users, but this is separate discussion haha

PR #719, opened 2021-11-03
refactor(tags): use dep injection for TagsService

### LINKED ISSUE #648: Refactor TagsService to use dependency injection

TagsService currently does not use dependency injection and instead imports models directly from `../../bootstrap/sequelize`. This will be troublesome when unit tests are being written.

Refactor code such that these imports:

```ts
import {
  Agency as AgencyModel,
  Post as PostModel,
  Tag as TagModel,
  User as UserModel,
} from '../../bootstrap/sequelize'
```

will be passed in as arguments when constructing the module instead.

**Additional context**

Similar issue mentioned in #646.

LoneRifle:
## Problem and Solution

Inject Sequelize models into TagsService, switching over to shared types as well

Closes #648

### DIFF
--- a/server/src/bootstrap/index.ts
+++ b/server/src/bootstrap/index.ts
@@ -162,7 +162,7 @@ const apiOptions = {
   tags: {
     controller: new TagsController({
       authService,
-      tagsService: new TagsService(),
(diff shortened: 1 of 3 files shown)


LinHuiqing (approved):
lgtm

Issue #649, opened 2021-10-25
Add unit tests for Tags module


LinHuiqing:
**Is your feature request related to a problem? Please describe.**

Unit tests are missing from the Tags module, meaning that it might be harder for us to track bugs when they occur.

**Describe the solution you'd like**

Unit tests for Tags module. 

**Additional context**

Blocked by #648.

PR #721, opened 2021-11-03
test(tags): add some test coverage for TagsService

LoneRifle:
## Problem and Solution

Provide unit tests just for listTags and getSingleTag, given
that we will change the relationship between tags, users and posts
in future iterations

Part of work for #649

### DIFF
--- a/server/src/modules/tags/__tests__/tags.service.spec.ts
+++ b/server/src/modules/tags/__tests__/tags.service.spec.ts
@@ -1,0 +1,132 @@
+import { Model, Sequelize } from 'sequelize'
+import {
+  Agency,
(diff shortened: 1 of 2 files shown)


LinHuiqing (approved):
lgtm
maybe mock sequelize and throw sequelize-related errors in the future?

LoneRifle:
not sure - we might want to completely remove the untested functions in TagsService given how we are moving to topics, and thus seeing tags as used by agency or user is somewhat meaningless

LinHuiqing:
Oh, I meant for within the tested functions, e.g. `await Tag.findOne` in `getSingleTag` might fail.

PR #725, opened 2021-11-05
build(deps): update validator to 13.7.0

LoneRifle:
## Problem and Solution

Update validator to 13.7.0 as part of routine updates to dependencies


LinHuiqing (approved):
lgtm

PR #727, opened 2021-11-05
build(deps): use @babel/eslint-parser

LoneRifle:
## Problem and Solution

Use @babel/eslint-parser for .js(x) files


LoneRifle:
Abandoning, more productive to just move to TypeScript

PR #741, opened 2021-11-08
fix(edit): use secondary.700 for confirm button

LoneRifle:
## Problem and Solution

Avoid using primary colours for the confirm button, since its red hues may confuse users into thinking it's a destructive action


wanlingt (approved):
lgtm!

PR #755, opened 2021-11-11
refactor(searchbox): darken shadow, use Chakra UnorderedList

LoneRifle:
## Problem and Solution

Darken the shadow around search box results to highlight it better to user.

Take the chance to use Chakra's unordered list instead of `<ul>`, ensuring we override the former's CSS margin to retain layout

### Before
Screenshot_20211111-112623

### After
image-1

### DIFF
--- a/client/src/components/SearchBox/SearchBox.styles.scss
+++ b/client/src/components/SearchBox/SearchBox.styles.scss
@@ -46,7 +46,7 @@
         z-index: 50;
         margin-inline-start: 0;
         // copied box-shadow from chakra UI shadow 2xl,
-        // with color from dark-lg
+        // with a much darker shade
         box-shadow: 0 25px 50px -12px rgba(0, 0, 0, 0.6);
         border-radius: $btn-radius; 
       }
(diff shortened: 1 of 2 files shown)


client/src/components/SearchBox/SearchBox.styles.scss
```diff
@@ -44,8 +44,10 @@
         width: 100%;
   
         z-index: 50;
-        // copied box-shadow from chakra UI shadow 2xl
-        box-shadow: 0 25px 50px -12px rgba(0, 0, 0, 0.25);
+        margin-inline-start: 0;
+        // copied box-shadow from chakra UI shadow 2xl,
+        // with color from dark-lg
+        box-shadow: 0 25px 50px -12px rgba(0, 0, 0, 0.6);
```
wanlingt, line 50: is the color from the dark-lg [here](https://github.com/chakra-ui/chakra-ui/blob/75edcf41e7ff4acc2569f2169949063c164d8f6e/packages/theme/src/foundations/shadows.ts)? those dark-lg values look slightly different to me

wanlingt (approved):
lgtm, i assume @khaleedah has approved the look of the shadow and colour

client/src/components/SearchBox/SearchBox.styles.scss
LoneRifle: Yea okay, they're not, I'll adjust the comment accordingly

PR #756, opened 2021-11-11
feat(search): add indexAllData controller and tests

### LINKED ISSUE #720: Set up search module to index database on OpenSearch
This is so that OpenSearch will be able to search data in our existing database. This can be done with [OpenSearch’s Node.js client SDK](https://github.com/opensearch-project/opensearch-js).

Before any search can be performed, OpenSearch has to index the database. This will be done by pulling relevant tables from the database, merging them into a single dataset, then adding them to the OpenSearch index with a bulk operation. While using a common index is not necessary for search to be performed, this reduces the complexity of search queries which would then improve the efficiency of searches. This new index (named ‘searchEntries’) will include details from both the post table and the answers table, as the search bar should find relevant entries based on values in these tables.
The following fields will then be used in the searchEntries index:
- title: string - content to search
- description: string | null - content to search
- answer: string - content to search
- agencyId: number - allows search filtered by agency
- postId: number - allows navigation to post
- topicId: number | null - allows search within topic

LinHuiqing:
## Problem

Add search controller and tests for indexAllData.

Closes #720

## Solution

Implement indexAllData method in controller and relevant tests.

### DIFF
--- a/server/src/modules/search/search.controller.ts
+++ b/server/src/modules/search/search.controller.ts
@@ -10,6 +10,11 @@ import { SearchEntry, SearchService } from './search.service'
 
 const logger = createLogger(module)
 
+/**
+ * Handles logic requiring other services, errors and returns responses accordingly.
+ * Note: Some methods will not return HTTP responses and instead return ResultAsync
+ * for internal server use e.g. indexAllData
+ */
 export class SearchController {
   private answersService: Pick<AnswersService, 'listAnswers'>
   private postService: Pick<PostService, 'listPosts'>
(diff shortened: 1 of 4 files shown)


server/src/modules/search/search.controller.ts
```diff
@@ -0,0 +1,96 @@
...
+import { SortType } from '../../types/sort-type'
+import { AnswersService } from '../answers/answers.service'
+import { DatabaseError } from '../core/core.errors'
+import { InvalidTagsError, InvalidTopicsError } from '../post/post.errors'
+import { PostService } from '../post/post.service'
+import { SearchEntry, SearchService } from './search.service'
+
+const logger = createLogger(module)
+
+export class SearchController {
```
LoneRifle, line 18: Could you include some jsdoc here to explain the nature of this controller?

LoneRifle (approved):
lgtm otherwise

PR #757, opened 2021-11-11
refactor(searchbox): migrate to Chakra

LoneRifle:
## Problem and Solution

- Migrate SearchBox to TypeScript and Chakra, taking advantage of theming to keep styling separate
- Take the opportunity to add button radius to the default theme
  - Point relevant CSS styles to use the new Chakra variable, and remove `misc.scss`


client/src/components/Header/Header.component.jsx
```diff
@@ -189,7 +189,7 @@ const Header = () => {
             maxW="680px"
             w="100%"
           >
-            <SearchBoxComponent agencyId={agency?.id} />
+            <SearchBox agencyId={agency?.id} />
```
LinHuiqing, line 192: Not this line, but for line # 290 of the same file, you forgot to change `SearchBoxComponent` to `SearchBox`

LinHuiqing (changes requested):

client/src/components/Header/Header.component.jsx
LinHuiqing: lgtm otherwise, but didn't want to approve in case this isn't addressed (it will break the app)

LinHuiqing (approved):

PR #762, opened 2021-11-12
refactor(login): migrate to ts, Chakra

LoneRifle:
## Problem and Solution

As part of an overall effort to depend more on Chakra,
migrate away from using manual styling and JavaScript to Chakra theming and TypeScript

## Tests
- Ensure login screen looks the same


client/src/pages/Login/Login.component.tsx
```diff
@@ -1,23 +1,30 @@
...
+        mx="auto"
+        maxW="calc(320px + 48px * 2)"
+        className="login-page"
+      >
         <Spacer h={['64px', '64px', '84px']} />
-        <h2>AskGov</h2>
-        <p>Answers from the Singapore Government</p>
+        <Box as="h2" textStyle="display-2" mb="8px">
+          AskGov
+        </Box>
```
LinHuiqing: `as="h2"` does not do anything. Also `Text` component might be a better component to use as this is just text. Something like:
```ts
<Text textStyle="display-2" mb="8px">
  AskGov
</Text>
```

client/src/pages/Login/Login.component.tsx
```diff
@@ -1,23 +1,30 @@
...
   if (user) {
     return <Navigate replace to={`/agency/${user.agency.shortname}`} />
   } else {
     return (
-      <div className="login-page">
+      <Box
+        mt="48px"
+        px={{ base: '36px', xs: '48px' }}
+        mx="auto"
+        maxW="calc(320px + 48px * 2)"
```
LinHuiqing, line 17: `"calc(...)"` should be changed to a constant since calculation would always be the same

client/src/pages/Login/Login.component.tsx
```diff
@@ -1,23 +1,30 @@
...
+        className="login-page"
```
LinHuiqing, line 18:
```suggestion

```
Unnecessary line^

LinHuiqing (changes requested):

client/src/pages/Login/Login.component.tsx
LoneRifle: Let's keep that for now - we might want to further refactor the theming to use a separate file

client/src/pages/Login/Login.component.tsx
LoneRifle: I feel that this is clearer, as it shows what we are looking for, ie, "320 pixels of width, with additional provisions for padding"

client/src/pages/Login/Login.component.tsx
LoneRifle: So something like this? ok.
```suggestion
        <Text textStyle="display-2" mb="8px">
          AskGov
        </Text>
```

LinHuiqing: yup 👍

LinHuiqing (approved):

PR #763, opened 2021-11-12
refactor(search): migrate to ts, Chakra

LoneRifle:
## Problem and Solution
As part of a migration to current norms, migrate SearchResults to use TypeScript and Chakra theming

### Before

image

image

### After

image

image


wanlingt (approved):
lgtm

PR #764, opened 2021-11-12
refactor: remove CSS-based breakpoints, no longer used

LoneRifle:
## Problem and Solution

Remove `styles/_sizes.scss` and all references, no longer used


LinHuiqing (approved):
lgtm

PR #765, opened 2021-11-12
refactor(links): migrate to TypeScript

LoneRifle:
## Problem and Solution

- Clean up LinkButton and PostQuestionButton, migrating both to TypeScript
- Delete LinkButton.styles.scss, never used


wanlingt (approved):
lgtm

Issue #688, opened 2021-10-28
Set up queries to OpenSearch engine on backend


LinHuiqing:
API: GET /search?agency=<agency>&query=<query>
Returns list of post titles sorted by score

PR #786, opened 2021-11-16
feat(search): add searchPosts to service and related tests

LinHuiqing:
## Problem

Related #688

## Solution

Set up search query (`searchPosts`) on search service.

**Features**:

- fulfills basic requirements:
    1. Able to search multiple fields (title, description and answer)
    2. Able to filter by agency
    3. Able to account for typos
- uses scenario-based testing (hence overlapping coverage)

### DIFF
--- a/server/src/modules/search/search.service.ts
+++ b/server/src/modules/search/search.service.ts
@@ -107,6 +107,13 @@ export class SearchService {
     })
   }
 
+  /**
+   * Searches posts in opensearch index
+   * @param index index name
+   * @param searchQuery search query entered by user
+   * @param agencyId agency id
+   * @returns results async with relevant search entries ranked by relevance
+   */
   searchPosts = async (
     index: string,
     searchQuery: string,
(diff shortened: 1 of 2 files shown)


server/src/modules/search/search.service.ts
```diff
@@ -103,4 +106,38 @@ export class SearchService {
       })
     })
   }
+
+  searchPosts = async (
+    index: string,
+    searchQuery: string,
+    agencyId?: number,
+  ) => {
+    const multiSearchType: QueryDslTextQueryType = 'most_fields'
```
LoneRifle: why not just hardcode this directly into `multiMatchQuery`?

LoneRifle (approved):
lgtm otherwise!

server/src/modules/search/search.service.ts
LinHuiqing: because the string needs to be typed haha but okay I can type `multiMatchQuery` directly too

Issue #746, opened 2021-11-09
Improve image upload feature


wanlingt:
**Is your feature request related to a problem? Please describe.**
The current image upload tool is not intuitive, and has the following problems:
- not obvious that alt text is required
- sizing is confusing

**Describe the solution you'd like**
Update the design of the image upload tool according to the new [Figma designs](https://www.figma.com/file/O6lxjjfn04Ow6nn2TMStpk/AskGov-Design-Master-v2?node-id=6271%3A69638)
- Prompt user to fill in alt text
- Different sizing options, e.g.
![image](https://user-images.githubusercontent.com/56983748/140884972-ed6cf38e-1909-4c83-8007-2804db80cc24.png)

PR #807, opened 2021-11-17
feat(image-upload): update image upload modal 

wanlingt:
## Problem

The current image upload tool is not intuitive.

Working towards #746 

## Solution

Create a custom image component for the rich text editor, referencing updated design of image modal in [Figma](https://www.figma.com/file/O6lxjjfn04Ow6nn2TMStpk/AskGov-Design-Master-v2?node-id=6271%3A69638)
Main changes include:
- Image preview, file name and size are shown for uploaded image
- Alt text input is displayed only if image source exists
- Spinner indicates loading image status
- Trash icon provided to remove image
- Image icon in RTE is only visible in desktop view
- URL upload option has been removed due to Content Security Policy (see comment below)

Additionally, Chakra theming is used to provide styles

## Before & After Screenshots

**BEFORE**:

https://user-images.githubusercontent.com/56983748/142226804-8c95e108-22f0-4695-b421-936575698541.mov

**AFTER**:

https://user-images.githubusercontent.com/56983748/142381616-915521bd-14b6-48a0-96ba-62cd5588fd71.mov

## Tests

- Insert image using file upload
- Modal should be cleared upon clicking "Cancel" or "Insert Image"
- Try to upload an image without keying in alt text - you should not be able to click the "Insert Image" button
- Inspect image, alt text should be whatever was entered


LoneRifle (changes requested):
lgtm otherwise. Couple of thoughts:

- Due to our strict Content Security Policy, we should not allow for embedding images directly by URL, since allowing arbitrarily-hosted images might open us up to potential issues where images hosted on an untrusted host can potentially exploit end-users through browser. Image hotlinking is also not good practice since it imposes a burden on the image host.
- Consider using Chakra theming (study SearchBox as an example)

wanlingt:
> lgtm otherwise. Couple of thoughts:
> 
> * Due to our strict Content Security Policy, we should not allow for embedding images directly by URL, since allowing arbitrarily-hosted images might open us up to potential issues where images hosted on an untrusted host can potentially exploit end-users through browser. Image hotlinking is also not good practice since it imposes a burden on the image host.
> * Consider using Chakra theming (study SearchBox as an example)

For embedding images by URL, would this be mitigated by only accepting URLs from file.go.gov.sg? Or would it better to just remove the URL option completely?

LoneRifle:
> For embedding images by URL, would this be mitigated by only accepting URLs from file.go.gov.sg? Or would it better to just remove the URL option completely?

That's a helpful mitigation! I would generally prefer that we host our own images though, since the trouble taken to upload a file on Go could have been spent just uploading into AskGov.

wanlingt:
> > > lgtm otherwise. Couple of thoughts:
> > > 
> > > * Due to our strict Content Security Policy, we should not allow for embedding images directly by URL, since allowing arbitrarily-hosted images might open us up to potential issues where images hosted on an untrusted host can potentially exploit end-users through browser. Image hotlinking is also not good practice since it imposes a burden on the image host.
> > > * Consider using Chakra theming (study SearchBox as an example)
> > 
> > 
> > For embedding images by URL, would this be mitigated by only accepting URLs from file.go.gov.sg? Or would it better to just remove the URL option completely?
> 
> That's a helpful mitigation! I would generally prefer that we host our own images though, since the trouble taken to upload a file on Go could have been spent just uploading into AskGov.

ok yup makes sense, unless agencies already use Go to generate image links - though I doubt this is very likely. Will hide the option for now, @khaleedah says to park it for sync discussion next week.

LoneRifle (approved):

LinHuiqing:
https://user-images.githubusercontent.com/37061143/142789673-91ecd363-703c-4546-8702-e7f0b10237a6.mov

Not sure if its intentional, but it's not very obvious the limits of where the user can drop their image. Might want to create an issue for this if it's not intentional.

wanlingt:
> Screen.Recording.2021-11-22.at.9.53.23.AM.mov 
> Not sure if its intentional, but it's not very obvious the limits of where the user can drop their image. Might want to create an issue for this if it's not intentional.

yeah i've noticed, thanks for spotting it too! fixing in a separate PR

PR #809, opened 2021-11-18
refactor(header): migrate bits to Chakra, tsx, tweak AgencyLogo

LoneRifle:
## Problem and Solution

As part of ongoing efforts to modernise the UI, migrate AgencyLogo and Banner
to TypeScript-based Chakra, and tweak the behaviour for AgencyLogo to ensure
the DOM remains stable until everything fully rendered.

- Consolidate Banner styling and layout into one Chakra-based tsx
- Move AgencyLogo to TypeScript
  - Allow `display` and margin-related props to be passed to the root
    Box of AgencyLogo, allowing us to remove a wrapper Box in Header,
    instead supplying those props directly to AgencyLogo
- Use `Header.agencyShortName` to determine if we need AgencyLogo,
  not `agency`. If specified, we should render the component used
  to show the agency logo even if `agency` is undefined and all we
  can show to the user is a placeholder. This keeps the DOM fairly
  stable while loading up everything for the user

## Screenshots

https://user-images.githubusercontent.com/10572368/142377775-1414d9c4-9e1f-473d-b3c1-c030390bebf3.mov


LinHuiqing (approved):
lgtm

PR #816, opened 2021-11-19
feat(search): add index backfill script and add index exist check before creating index

LinHuiqing:
## Problem

Backfill of opensearch was previously in the `search` module itself with the intention of running it being executed when a server instance starting up. However, this is not ideal as it will cause concurrency issues when multiple instances of the server is spun up on AWS Elastic Beanstalk.

## Solution

Extract code to backfill opensearch index to a submodule which can be triggered by a script.

**Features**:

- extract backfill code into a submodule
- trigger opensearch indexing when backfill script is run
- check if index exists before creating it

**Improvements**:

- fix error typing in search using ResponseError
- modify routes to use relative path /
- test suite modified accordingly
- remove `testPathIgnorePatterns` config from jest

## Test

1. `npm run dev`
2. On a separate terminal, `cd server/src/bootstrap && npx ts-node search-backfill-trigger.ts`
3. `curl -HEAD 'https://localhost:9200/search_entries' --insecure -u 'admin:admin'` should return a valid response now


LoneRifle (approved):
lgtm. I realise you follow up this PR with #897, and will look at that in due course.

PR #833, opened 2021-11-22
refactor: move remaining js(x) files to TypeScript

LoneRifle:
## Problem

Because we still have vanilla JavaScript files, we still have to keep babel-eslint,
which is holding us back from the eslint v8

## Solution

- Convert all remaining js/jsx files to TypeScript, adjusting
  for types in the process
- Remove references to React global, using proper imports instead
- Configure eslint to recognise the JSX global and ignore the
  react-scripts config overrides


wanlingt (approved):
lgtm

PR #835, opened 2021-11-22
feat(search): expose search query endpoint

### LINKED ISSUE #688: Set up queries to OpenSearch engine on backend

LinHuiqing:
- use relative paths for routes within search module
- mount search routes on app

Closes #688

Note: As data has not been indexed on opensearch and my mistake of creating custom errors, the generic 'Internal Server Error' message will be returned when the search endpoint is hit. This will be fixed in PR #816. For now, opensearch errors can be viewed on [CloudWatch](https://ap-southeast-1.console.aws.amazon.com/cloudwatch/home?region=ap-southeast-1#logsV2:log-groups/log-group/$252Faws$252Felasticbeanstalk$252Fhelpgovsg-staging-env$252Fvar$252Flog$252Feb-docker$252Fcontainers$252Feb-current-app$252Fstdouterr.log).

## Tests

Send a GET request to `/api/v1/search?query=hi`. Return body should be `{"message":"Internal Server Error"}`. 

If not mounted properly, return body will be the standard `/not-found` response, which is the standard `index.html` page for built apps and the following if in development environment:

```html
<!DOCTYPE html>
<html lang="en">

<head>
	<meta charset="utf-8">
	<title>Error</title>
</head>

<body>
	<pre>Cannot GET /api/v1/search</pre>
</body>

</html>
```

### DIFF
--- a/server/src/bootstrap/index.ts
+++ b/server/src/bootstrap/index.ts
@@ -62,6 +62,10 @@ import {
   User,
 } from './sequelize'
 import sessionMiddleware from './session'
+import { SearchController } from '../modules/search/search.controller'
+import { SearchService } from '../modules/search/search.service'
+import { searchConfig } from './config/search'
+import { Client } from '@opensearch-project/opensearch'
 
 export { sequelize } from './sequelize'
 export const app = express()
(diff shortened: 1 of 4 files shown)


LinHuiqing:
By the way, this is up on staging to try out now too

wanlingt (commented):
lgtm, have checked that test works and is reflected on cloudwatch. Though I'm not too sure about the certificate part so will leave that for @LoneRifle to approve / comment on

LinHuiqing:
Ah, the certificate part is for the client to connect to opensearch engine. So if it works on staging it should be good. I should probably add a comment for how we got to that path though, let me add that in. Thanks!

wanlingt (approved):

server/src/bootstrap/index.ts
```diff
@@ -128,6 +132,31 @@ const answersService = new AnswersService({ Post, Answer })
...
+  connectionOptions = { rejectUnauthorized: false } // Turn off certificate verification (rejectUnauthorized: false)
+} else {
+  connectionOptions = {
+    ca: fs.readFileSync('/etc/ssl/cert.pem'), // Path to CA certificate for Alpine Linux, which is the OS AskGov runs on
+  }
+}
+const client = new Client({
+  node: protocol + '://' + auth + '@' + searchHost + ':' + port,
+  ssl: connectionOptions,
+})
```
LoneRifle, line 152: I would put this into `bootstrap/search.ts`, similar to how we bootstrap the Sequelize client. Could you file a PR for this please?

LinHuiqing: Ah sure 👍

PR #848, opened 2021-11-23
refactor(search): extract search client setup to a separate file

LinHuiqing:
## Problem

Spaghetti code in bootstrap index should be a no-no.

Resolves https://github.com/opengovsg/askgovsg/pull/835#discussion_r754785683

## Solution

- for better modularity, based on `bootstrap/sequelize`

### DIFF
--- a/server/src/bootstrap/search.ts
+++ b/server/src/bootstrap/search.ts
@@ -1,0 +1,27 @@
+import { Client } from '@opensearch-project/opensearch'
+import fs from 'fs'
+import { baseConfig, Environment } from './config/base'
+import { searchConfig } from './config/search'
+
+const searchHost = searchConfig.host
+const protocol = 'https'
+const port = searchConfig.port
+const auth = `${searchConfig.username}:${encodeURIComponent(
+  searchConfig.password,
+)}`
+
+const ssl =
+  baseConfig.nodeEnv === Environment.Dev
+    ? // Turn off certificate verification (rejectUnauthorized: false)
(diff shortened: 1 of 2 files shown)


server/src/bootstrap/search.ts
```diff
@@ -0,0 +1,27 @@
...
+} else {
+  connectionOptions = {
+    ca: fs.readFileSync('/etc/ssl/cert.pem'), // Path to CA certificate for Alpine Linux, which is the OS AskGov runs on
+  }
+}
+
+export const searchClient = new Client({
+  node: protocol + '://' + auth + '@' + searchHost + ':' + port,
+  ssl: connectionOptions,
+})
```
LoneRifle: So what about...
```suggestion
const ssl = baseConfig.nodeEnv === Environment.Dev
  // Turn off certificate verification (rejectUnauthorized: false) 
  // if we are running in a dev environment
  ? { rejectUnauthorized: false } 
  // Otherwise, set the path to CA certificate for Alpine Linux, 
  // the OS that AskGov runs on
  : { ca: fs.readFileSync('/etc/ssl/cert.pem') } 

export const searchClient = new Client({
  node: `${protocol}://${auth}@${searchHost}:${port}`,
  ssl,
})
```

LoneRifle (approved):
lgtm otherwise

server/src/bootstrap/search.ts
LinHuiqing: okay done, thanks!

PR #853, opened 2021-11-24
fix(post): fix getRelatedPosts sequelize query

wanlingt:
## Problem

getRelatedPosts does not follow the intended ordering of posts, namely that posts with the same topicId are not displayed at the top.

## Solution

Instead of `COUNT`, use a `CASE WHEN` to identity posts with the same topicId

## Before & After Screenshots

**BEFORE**:

Screenshot 2021-11-24 at 12 49 28 PM
**AFTER**:
Screenshot 2021-11-24 at 12 51 44 PM

### DIFF
--- a/server/src/modules/post/post.service.ts
+++ b/server/src/modules/post/post.service.ts
@@ -128,14 +128,16 @@ export class PostService {
           attributes: {
             include: [
               [
-                Sequelize.fn('COUNT', Sequelize.col('topicId')),
-                'relatedTopics',
+                Sequelize.literal(
+                  `CASE WHEN TopicId = ${post.topicId} THEN 1 ELSE 0 END`,
+                ),
+                'relatedTopic',
               ],
               [Sequelize.fn('COUNT', Sequelize.col('tags.id')), 'relatedTags'],
             ] as ProjectionAlias[],
           },
           order: [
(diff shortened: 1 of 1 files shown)


server/src/modules/post/post.service.ts
```diff
@@ -128,14 +128,16 @@ export class PostService {
           attributes: {
             include: [
               [
-                Sequelize.fn('COUNT', Sequelize.col('topicId')),
-                'relatedTopics',
+                Sequelize.literal(
+                  `CASE WHEN TopicId = ${post.topicId} THEN 1 ELSE 0 END`,
```
LoneRifle, line 132:
```suggestion
                  `CASE WHEN topicId = ${post.topicId} THEN 1 ELSE 0 END`,
```
Not very comfortable with this, partly because of a very small possibility of SQL injection happening by messing with the topicId value. Would it make more sense to adjust the `where` clause to take topic IDs into account?

LoneRifle (approved):
Lgtm otherwise, as discussed concerning biz requirements

server/src/modules/post/post.service.ts
wanlingt: To document what was discussed - the intent is to grab the most popular questions if we can't find posts with the same topicId. Related questions implementation to be made more robust once we have text search capabilities

PR #857, opened 2021-11-24
fix(posts): limit related posts to the same agency

wanlingt:
## Problem

Currently, related posts do not necessarily belong to the same agency. This does not make much sense as each agency's content is quite siloed.

## Solution

Add a `where` clause in the sequelize query to filter for related posts belonging to the same agency


LoneRifle (approved):
Lgtm for now, we need something else as we transition into something more like Quora for government

PR #858, opened 2021-11-24
refactor(auth-form): migrate to Chakra theme, adjusting layout

LoneRifle:
## Problem and Solution

- Move AuthForm to use Chakra with theming, removing one more dependency on CSS-based text styles
- Take the opportunity to remove one spurious import from RichTextEditor.styles.scss


wanlingt (commented):
Left - develop, Right - staging
image

styling looks slightly different..

wanlingt (approved):
lgtm!

PR #889, opened 2021-12-01
refactor(permissions): remove permissions

### LINKED ISSUE #859: Remove Permissions type, Sequelize model
The Permissions type is no longer needed as we now control access to posts depending on user's agency. The Permissions type, Sequelize model and references to both should hence be removed from the codebase, with corresponding changes made to the database in prod

wanlingt:
## Problem

The Permissions type and Sequelize model are no longer needed as we now control access to posts depending on a user's agency.

Closes #859 

## Solution
- Remove all references to the Permissions type and Sequelize model
- Rename `defineUserAndPermissions` to `defineUser`

## Tests

- All actions relating to user permissions (login, modification of posts) should work as before

## Deploy Notes

**New scripts**:

- `20211201101858-remove-permissions.js`: remove permissions Sequelize table


client/src/contexts/AuthContext.tsx
```diff
@@ -7,7 +7,7 @@ import { LoadUserDto } from '~shared/types/api'
 import { useStyledToast } from '../components/StyledToast/StyledToast'
 
 interface AuthContextProps {
-  user: LoadUserDto | null
+  user: LoadUserDto
```
LoneRifle, line 10: This should be nullable, since the response from the backend could possibly be null

LoneRifle (approved):
lgtm otherwise

client/src/contexts/AuthContext.tsx
wanlingt: LoadUserDto is already nullable

`export type LoadUserDto =
  | (User & {
      agency: Agency
    })
  | null`,

hence I removed the `null` here. Would that make sense?

LoneRifle: Ah, true. Okay then!

PR #897, opened 2021-12-03
feat(search): set doc _id as postId to access search entries with postId easily

LinHuiqing:
## Problem

As services to post and answer tables are separate, operations to add and update posts on the opensearch index have to be able to accommodate having questions and answers added separately too. In addition, doing so would allow for easier implementation of having questions and answers added by members of the public.

## Solution

- Set document _id as postId for easy access of search entries
- Change `answer` field to `answers` and take in list of answers. 

## Tests

Send a search query which will give results (e.g. "/search?query=What"). In the results returned in the response body, we should observe that for the search results:
1. `_id` should be the same as their `postId` 
2. `answers` should contain a list of strings

### DIFF
--- a/server/src/modules/search/__tests__/search.controller.spec.ts
+++ b/server/src/modules/search/__tests__/search.controller.spec.ts
@@ -123,7 +123,7 @@ describe('SearchController', () => {
           id: 3,
           options: {},
           params: {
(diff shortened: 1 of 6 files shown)


LoneRifle (approved):
lgtm

Issue #687, opened 2021-10-28
Synchronise Opensearch Index with database


LinHuiqing:
This is so that Elasticsearch will be able to search data in our existing database. Refer to [Logstash documentation](https://www.elastic.co/guide/en/logstash/current/index.html) and relevant plugin(s) which can help ingest data from Sequelize to be used in Elasticsearch. [Jdbc input plugin](https://www.elastic.co/guide/en/logstash/current/plugins-inputs-jdbc.html) seems to be a candidate.
Try the integration on staging.

PR #923, opened 2021-12-08
feat(search): add createPost and updatePost for search entries

LinHuiqing:
## Problem

No infrastructure to add and update a search entry to synchronise opensearch index with database when posts are added to the database.

Working towards #687

## Solution

Add sync service to include createPost and updatePost, which is called by postService to synchronise opensearch's index with the database. Includes relevant unit tests as well.

## Tests

Setup:

1. Login as an administrator.

3. Create a post. A POST request to `/posts` should be observed in the network panel as usual, as seen in the screenshot:

    Screenshot 2021-12-13 at 2 08 42 PM

    This means that the POST request to add the post was successful. Payload included in the request should be something like the one below. Take note of the values in the `title` and `description` fields:

    Screenshot 2021-12-13 at 2 08 50 PM

    The doc `_id`, which is set to the post id, can be seen in the response, which is **44** in this example as seen below.

    Screenshot 2021-12-13 at 2 09 01 PM

    This value should match the post id in the URL.

4. Check that the post has been indexed on opensearch with a GET request `curl -XGET 'https://localhost:9200/search_entries/_doc/44' --insecure -u 'admin:admin'`. `_version` should be 1 as the post is freshly added.

    Screenshot 2021-12-13 at 2 09 36 PM

5. Edit the post. A PUT request to `/post/<post_id>` should be observed in the network panel as usual, as seen in the screenshot.

    Screenshot 2021-12-13 at 2 19 12 PM

    This means that the PUT request to add the post was successful. Payload included in the request should be something like the one below. Take note of the updated values in the `title` and `description` fields:

    Screenshot 2021-12-13 at 2 19 53 PM

    The response should say that the answer has been updated.

    Screenshot 2021-12-13 at 2 19 26 PM

6. Again, check that the post has been updated on opensearch with a GET request `curl -XGET 'https://localhost:9200/search_entries/_doc/44' --insecure -u 'admin:admin'`. `_version` should be 2 as the entry has been changed once after creation.

    Screenshot 2021-12-13 at 2 20 07 PM

To simulate errors being thrown by opensearch's API, simply change code in the sync service (`server/src/modules/search/sync/sync.service.ts`) such that it will fail. For instance, `this.client.create` can be changed to `this.client.update` and vice versa. Rebuilding the file after these code changes is not required as the server is hot-reloaded.

7. Create a post. The post creation should fail with status code `500 Internal Server Error`.

    Screenshot 2021-12-13 at 2 33 04 PM

    Post should not show up on the database or opensearch index as well.

8. Update a post. Similar to the previous point, the request should fail with `500 Internal Server Error` and fields which were edited should not be updated on both the database and opensearch index.


LinHuiqing:
Sorry, just realised it should have private access. (edit: as discussed, code should be refactored such that syncing is done within a sequelize transaction)

server/src/modules/search/search.types.ts
```diff
@@ -0,0 +1,8 @@
+export type SearchEntry = {
+  postId: number
+  title?: string
+  description?: string | null
+  answers?: string[]
+  agencyId?: number
+  topicId?: number | null
+}
```
LoneRifle: Should this belong to `shared/`?

LoneRifle (approved):
lgtm otherwise

server/src/modules/search/search.types.ts
LinHuiqing: Hm yes, it would come it handy when the query is hooked up to frontend. Will add it in `shared/` then, thanks!

PR #924, opened 2021-12-08
refactor(post): migrate to Chakra theme

LoneRifle:
## Problem and Solution

Move all styling information from Post and companion CSS file to
equivalent Chakra theme, removing spurious divs in the process

- Create Chakra theme for Post
- Move styling information from component and CSS to new theme
- Remove post-side-with-edit and question-main divs, flattening
  children into parent


LinHuiqing (approved):
lgtm

PR #926, opened 2021-12-09
refactor(post-item): move to Chakra theme

LoneRifle:
## Problem and Solution

Move PostItem to Chakra theming, taking the trouble to simplify its DOM

- Migrate theming from component and scss to PostItem Chakra theme
- Flatten as much of the DOM in PostItem as possible


client/src/components/PostItem/PostItem.component.tsx
```diff
@@ -12,34 +11,25 @@ const PostItem = ({
...
+      {/* Title display area */}
+      <Link as={RouterLink} to={`/questions/${id}`}>
+        <Text sx={styles.linkText}>{title}</Text>
+      </Link>
+      {/* <Box sx={styles.description}>
             {description && <RichTextFrontPreview value={description} />}
-          </div> */}
-        </div>
-      </div>
+          </Box> */}
```
LinHuiqing, line 26: Is there a reason why this commented block is left in the file?

LinHuiqing (approved):
lgtm otherwise

client/src/components/PostItem/PostItem.component.tsx
LoneRifle: just in case we change our minds about having description

PR #943, opened 2021-12-13
feat(search): add updatePost to answer service

LinHuiqing:
## Problem

Answers are not indexed when posts are added or updated.

Working towards #687

## Solution

Update search entry with answers when answers are created or updated.

- uses updatePost for both createAnswer and updateAnswer as search entry would have been created by
createPost
- includes tests

## Tests

1. Login as an administrator.

3. Create a post. A POST request to `/posts/answers/<postId>` should be observed in the network panel as usual, as seen in the screenshot:

    Screenshot 2021-12-13 at 5 52 18 PM

    The answer should be sent in the payload as seen in the screenshot below:

    Screenshot 2021-12-13 at 5 52 49 PM

    The response is the answer id.

    Screenshot 2021-12-13 at 5 52 58 PM

4. Check that the answer in the post has been indexed on opensearch with a GET request `curl -XGET 'https://localhost:9200/search_entries/_doc/47' --insecure -u 'admin:admin'`. `_version` should be 2 as the search entry is first created when created by the post service, then updated when the answer is created.

    Screenshot 2021-12-13 at 5 53 39 PM

5. Edit the post. A PUT request to `/posts/answers/<answer_id>` should be observed in the network panel as usual, as seen in the screenshot.

    Screenshot 2021-12-13 at 5 57 02 PM

    This means that the PUT request to add the post was successful. Payload included in the request should be something like the one below. Take note of the value in the `text` field which contains the updated body of the answer:

    Screenshot 2021-12-13 at 5 57 11 PM

6. Check that the answer in the post has been indexed on opensearch with a GET request curl -XGET 'https://localhost:9200/search_entries/_doc/47' --insecure -u 'admin:admin'.

    Screenshot 2021-12-13 at 5 58 15 PM

7. To test for failure, we need to ensure that the sequelize transactions in post service do not cause the operation to fail before the answer service does so. As such, comment out the sync operations in `post.service`. As an example, the code below is commented out for `updatePost`. This should be done for `createPost` as well.

```ts
// const updatePostResponse = await this.searchSyncService.updatePost(
//   this.searchIndexName,
//   {
//     ...sharedObjValues,
//     postId: id,
//   },
// )
// if (updatePostResponse.isErr()) throw updatePostResponse.error
```

8. Next, in `sync.service`'s `updatePost`, replace `this.client.update` with `this.client.create`. This should cause opensearch's API to throw an error as the search entry should have been already been created.

9. Update a post. The answer update request should fail with status code `500 Internal Server Error`.

    Screenshot 2021-12-13 at 6 37 07 PM

    The request can be verified with the payload panel.

    Screenshot 2021-12-13 at 6 37 16 PM

     Verify that the request did not make any changed to opensearch's index by sending a GET request:

    Screenshot 2021-12-13 at 6 37 43 PM

    Verify that the request did not update the answer as well by checking the database or navigating to `/questions/<post_id>`.

10. Create a post. The answer creation request should fail with status code `500 Internal Server Error` as well.

### DIFF
--- a/server/src/bootstrap/index.ts
+++ b/server/src/bootstrap/index.ts
@@ -128,7 +128,12 @@ const postService = new PostService({
 })
 const enquiryService = new EnquiryService({ Agency, mailService })
 const recaptchaService = new RecaptchaService({ axios, ...recaptchaConfig })
-const answersService = new AnswersService({ Post, Answer })
+const answersService = new AnswersService({
+  Post,
+  Answer,
+  searchSyncService,
+  sequelize,
+})
 const topicsService = new TopicsService({ Topic })
 const userService = new UserService({ User, Tag, Agency })
 
--- a/server/src/bootstrap/search-backfill-trigger.ts
+++ b/server/src/bootstrap/search-backfill-trigger.ts
@@ -17,8 +17,13 @@ const triggerIndexAllData = async (searchController: BackfillController) => {
 // To delete search_entries index locally: curl -XDELETE 'https://localhost:9200/search_entries' --insecure -u 'admin:admin'
 
 if (require.main === module) {
-  const answersService = new AnswersService({ Post, Answer })
   const searchSyncService = new SyncService({ client: searchClient })
+  const answersService = new AnswersService({
+    Post,
+    Answer,
+    searchSyncService,
+    sequelize,
+  })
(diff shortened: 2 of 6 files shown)


LoneRifle (approved):
lgtm!

PR #947, opened 2021-12-14
feat(search/sync): sync opensearch index with db for deletePost operations

### LINKED ISSUE #687: Synchronise Opensearch Index with database

LinHuiqing:
## Problem

Search entries on opensearch would remain in index even when deleted on database.

Closes #687

## Solution

Do an API call to opensearch to delete relevant search entry when a post is deleted from the database.

- add deletePost to sync service
- call `sync.service`'s `deletePost` in `post.service`'s `deletePost`

## Tests

1. Login as an administrator.

3. Navigate to a post. The post used in the following screenshots is the one at `/questions/39`.
4. Check that the post is indexed on opensearch with `curl --insecure -u 'admin:admin' -XGET 'https://localhost:9200/search_entries/_doc/39'`.

    Screenshot 2021-12-14 at 1 01 11 PM

    If the post is not indexed on opensearch:
    1. Delete index with `curl --insecure -u 'admin:admin' -XDELETE 'https://localhost:9200/search_entries/'`
    2. Backfill opensearch index with data by navigating to `server/src/bootstrap` then running `npx ts-node search-backfill-trigger.ts`

5. Delete the post. The request should succeed as shown in the request on the network panel.

    Screenshot 2021-12-14 at 1 00 51 PM

6. Check that the post no longer exists on both database and opensearch's index. To check that it is no longer in database, try navigating to the post's page or check the MySQL db directly.

    To check that it is no longer in opensearch's index, send a GET request for it with `curl --insecure -u 'admin:admin' -XGET 'https://localhost:9200/search_entries/_doc/39'`

    Screenshot 2021-12-14 at 1 01 21 PM

To cause `sync.service` to throw an error, change the `index` field to a string which is not 'search_entries'.

7. Delete the post. The operation should fail as seen in the network panel:

    Screenshot 2021-12-14 at 1 32 56 PM

8. Check that the post still exists in db and opensearch index.

### DIFF
--- a/server/src/modules/post/__tests__/post.routes.spec.ts
+++ b/server/src/modules/post/__tests__/post.routes.spec.ts
@@ -68,6 +68,7 @@ describe('/posts', () => {
   const searchSyncService = {
     createPost: jest.fn(),
     updatePost: jest.fn(),
+    deletePost: jest.fn(),
   }
 
   beforeAll(async () => {
@@ -239,6 +240,7 @@ describe('/posts', () => {
     it('returns 200 if post is deleted', async () => {
       const { id } = mockPosts[0]
       authService.hasPermissionToAnswer.mockResolvedValue(true)
+      searchSyncService.deletePost.mockResolvedValue(okAsync({}))
 
       const path = '/posts'
       const app = express()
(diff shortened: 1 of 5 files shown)


LoneRifle (approved):
lgtm!

PR #962, opened 2021-12-15
feat(search/sync): replace fusejs search with /search

### LINKED ISSUE #690: Replace current search mechanism and decommission FuseJS

LinHuiqing:
## Problem

Search on frontend was still using fuse.js, which is limited to fuzzy search.

Closes #690

## Solution

Port search used on frontend over to use search queries powered by opensearch.

**Features**:

- Port search on search bar over to use `/search`
- Port search on search results page over to use `/search`

**Bug Fixes**:

- Change response format of `/search` to be `SearchEntry[]`

## Tests

1. Use the search bar on the main AskGov page and agency pages to make sure that only relevant search results show up.
2. Press enter in search bar to get to the search results page. Ensure that only the relevant search results show up.

## Deploy Notes

Removed `fuse.js` from client dependencies.

### DIFF
--- a/client/src/components/PostItem/PostItem.component.tsx
+++ b/client/src/components/PostItem/PostItem.component.tsx
@@ -8,7 +8,7 @@ import EditButton from '../EditButton/EditButton.component'
 const PostItem = ({
   post: { id, title, tags, agencyId },
 }: {
-  post: BasePostDto
+  post: Pick<BasePostDto, 'id' | 'title' | 'tags' | 'agencyId'>
 }): JSX.Element => {
   const { user } = useAuth()
   const styles = useMultiStyleConfig('PostItem', {})
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -21,7 +21,6 @@
         "downshift": "^6.1.7",
         "draft-js": "^0.11.7",
         "draftjs-to-html": "^0.9.1",
-        "fuse.js": "^6.4.6",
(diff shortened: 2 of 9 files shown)


client/src/components/PostItem/PostItem.component.tsx
```diff
@@ -8,7 +8,7 @@ import EditButton from '../EditButton/EditButton.component'
 const PostItem = ({
   post: { id, title, tags, agencyId },
 }: {
-  post: BasePostDto
+  post: BasePostDto | { id: number; title: string; tags: []; agencyId: number }
```
LoneRifle: what's going on here?

LoneRifle (approved):
lgtm otherwise, well done on completing this!

client/src/components/PostItem/PostItem.component.tsx
LinHuiqing: Because the search results returned by the search query is able to provide the `id`, `title` and `agencyId` but not the other properties in `BasePostDto`. However, an object with `BasePostDto` is passed into `PostItem` from `PostList`. Not sure how to do a 'minimal keys' thing so I just modified it to accept both types. 

Or would it be better to modify the object passed in from `PostList`?

LinHuiqing: As discussed, we'll use `Pick` :)

Issue #650, opened 2021-10-25
Add unit tests for web routes


LinHuiqing:
**Is your feature request related to a problem? Please describe.**

Web routes doesn't have unit tests, meaning that it might be harder to catch bugs when they occur.

**Describe the solution you'd like**

Add unit tests for web routes

**Additional context**

Continuation of #611, blocked by #646.

PR #966, opened 2021-12-16
test(web-routes): add unit tests to cover routes for web module

christopherlim98:
## Problem

Add unit tests to cover web routes, so that it is easier to detect any breaking changes relating to web routes in the future

Closes [#650]

## Solution

- Write unit tests for web routes
- Use `neverthrow` library


LoneRifle (approved):
lgtm

PR #989, opened 2021-12-20
docs(readme): add instructions to backfill index

christopherlim98:
Add instructions to readme on how to backfill index for search_entires, which is needed for OpenSearch integration.


LoneRifle (approved):
lgtm

PR #1008, opened 2021-12-22
chore(search): use integer type instead of long type when indexing

### LINKED ISSUE #945: Specify properties of fields when backfilling opensearch index with data

Fields indexed on OpenSearch currently have very loosely defined properties, i.e. numbers like ids are specified to be of `float` type, when the `integer` type would be enough. Specifying types which will help to [improve indexing and searching efficiency](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/number.html).

Specify field types as properties when backfilling opensearch.

**Additional context**

To check for properties of fields, simply use `curl --insecure -u 'admin:admin' -XGET 'https://localhost:9200/search_entries/'`.

Screenshot 2021-12-14 at 10 17 59 AM

christopherlim98:
## Problem

By default, OpenSearch builds the index using loosely defined properties. For example, `id` is marked with type `float` instead of `integer`. Specifying types which will help to[ improve indexing and searching efficiency.](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/number.html)

Closes #945

## Solution

Added `mapping` to `indexConfig` file. Note that [mapping cannot be changed for an existing field](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-put-mapping.html#updating-field-mappings). This means that applying the changes dynamically is slightly more complex. Look at the improvements section.

## Improvements

There are a few ways that we can change mapping without any downtime. 

- Method 1: reindex -> delete old index -> use alias on new index.
OpenSearch Index-Page-1 drawio

- Method 2: using alias from the get-go for the main search index
OpenSearch Index-Page-2 drawio (2)

Method 2 would require a refactor of how we currently reference the `search_entries` index in the post and answer services (assuming we do not delete the existing search_entries index).

**Pros of using aliases**
- Using aliases makes future index changes more reliable. We do not need to delete the existing index before running the backfill script. We can create the new index and reference it by the active alias name, and revert to the old index in case of failure.

**Cons of using aliases**
- Managing aliases may be messy. For e.g., we are not able to delete indices by referencing their alias. Although, it is still easy to check the alias mappings by `GET /_alias`.
- As multiple indices could occur at the same time, we need to (1) ensure there are no stale indices and (2) have an index naming strategy. Maybe new indices could have the date appended to the end of the index, for e.g. `search_entries_221221`.

**Further question**
- Should we write a bash script to document which CURL commands to handle slightly more complex operations, like reindexing or changing index setting dynamically (see #996 )? 

## Before & After Screenshots

**BEFORE**:
Screenshot 2021-12-22 at 1 38 46 PM

**AFTER**:
Screenshot 2021-12-22 at 1 33 43 PM

## Deploy Notes
Assuming we go ahead with method 1, it should look like

```sh
#----------- CHANGE INDEX MAPPING -----------#
# Create new index
curl -X PUT "https://localhost:9200/search_entries_copy" --insecure -u 'admin:admin' -H 'Content-Type: application/json' -d'
{
  "settings": {
    "analysis" : {
      "analyzer":{
        "default":{
          "tokenizer":"whitespace",
          "filter": ["stop_words_filter"]
        }
      },
      "filter" : {
        "stop_words_filter": {
          "type": "stop",
          "ignore_case" : true
        }
      }
    }
  }, 
  "mappings": {
    "properties" : {
      "agencyId" : {
        "type" : "integer"
      },
      "postId" : {
        "type" : "integer"
      },
      "topicId":{
        "type" : "integer"
      }
    }
  } 
}
'

# Reindex from old to new index
curl -X POST "https://localhost:9200/_reindex?pretty" --insecure -u 'admin:admin' -H 'Content-Type: application/json' -d'
{
  "source": {
    "index": "search_entries"
  },
  "dest": {
    "index": "search_entries_copy"
  }
}
'

# Delete old index
curl -X DELETE 'https://localhost:9200/search_entries' --insecure -u 'admin:admin'

# Add alias
curl -X POST "https://localhost:9200/_aliases?pretty" --insecure -u 'admin:admin' -H 'Content-Type: application/json' -d'
{
  "actions": [
    {
      "add": {
        "index": "search_entries_copy",
        "alias": "search_entries"
      }
    }
  ]
}
'
```

## Logistics
This PR has been blocked by #996; once that PR is merged to develop, this PR should be rebased on this branch.

**References**:
https://www.elastic.co/blog/changing-mapping-with-zero-downtime

### DIFF
--- a/server/src/modules/search/backfill/index-config.ts
+++ b/server/src/modules/search/backfill/index-config.ts
@@ -16,4 +16,17 @@ export const indexConfig = {
       },
     },
   },
+  mappings: {
+    properties: {
+      agencyId: {
+        type: 'integer',
+      },
+      postId: {
+        type: 'integer',
+      },
+      topicId: {
+        type: 'integer',
+      },
+    },
+  },
 }


LoneRifle (changes requested):
I don't think this change is necessary for a couple of reasons:
- storage is cheap even if `long` is twice as long as `integer`, esp given we don't have that many identifiers, and at any rate Elasticsearch (and by extension OpenSearch) is smart enough to deal with storage
- we don't search on ids anyway
- I would generally be more comfortable about using 64-bit identifiers where convenient to do so since wraparound wouldn't happen until _much_ later relative to 32-bit ones
- this is too much planning for premature optimisation that I doubt that we would need now

Close the PR at your leisure, or comment to continue the discussion

christopherlim98:
> I don't think this change is necessary for a couple of reasons:
> 
> * storage is cheap even if `long` is twice as long as `integer`, esp given we don't have that many identifiers, and at any rate Elasticsearch (and by extension OpenSearch) is smart enough to deal with storage
> * we don't search on ids anyway
> * I would generally be more comfortable about using 64-bit identifiers where convenient to do so since wraparound wouldn't happen until _much_ later relative to 32-bit ones
> * this is too much planning for premature optimisation that I doubt that we would need now
> 
> Close the PR at your leisure, or comment to continue the discussion

Understood, I agree with the mentioned points regarding the type specification.

Regarding the aliasing strategy, should that be documented/pursued in a separate thread at a later time? I think it might be helpful in the future if we ever need to rebuild the index. Although, it might be an over-optimisation if the current indexing step does not take too much time given the volume.

LoneRifle:
Sure, document the aliasing strategy in our internal documentation. Be sure to include references to any external articles you have.

christopherlim98:
Mentioned aliasing strategy with references in OpenSearch design doc. Closing PR.

PR #1020, opened 2021-12-25
fix(backend port number): change backend port number to 6174

### LINKED ISSUE #992: Port 5000 already in use by AirPlay on macOS 12
**Description**
Port 5000 used as the server port in the project is already in use by AirPlay on macOS 12. As such, when I run `npm run dev`, this the following error message is shown:
[ERROR] 14:44:54 Error: listen EADDRINUSE: address already in use :::5000
image

**Additional context**
Brief Googling reveals that this is an issue with Apple using port 5000 for AirPlay Receiver. See [here](https://developer.apple.com/forums/thread/682332) and [here](https://www.reddit.com/r/webdev/comments/qg8yt9/apple_took_over_port_5000_in_the_latest_macos/). 

I don't fully understand this, but apparently this is [not an unreasonable decision](https://www.reddit.com/r/webdev/comments/qg8yt9/comment/hi65w7i/):
> Just did a bit more digging, and IANA actually assigns port 5000 to "commplex-main" - so really any other service that's using that port is violating the official assignment. Based on a quick search of what commplex-main is, it's related to UPnP for network device discoverability. I'm not sure exactly what the AirPlay Receiver service does, but I get the gist that it probably does fit somewhere into the scope of UPnP-related features. At the very least, it's probably a far more accurate usage of port 5000 than some random dev servers.

**Solution**
A quick way to fix the problem is to disable AirPlay Receiver in System Settings > Sharing. 

However, I was thinking a better solution might be to simply use a different port number for the server. If I'm not mistaken, we can just use another port number that is less likely to result in such conflicts? If so, happy to submit a PR to change all instances of 5000 to, say, another randomly chosen port number.

zxt-tzx:
Current port number of 5000 clashes with Apple using port 5000 for AirPlay Receiver on Mac.

fixes #992

Notes:
- Port number of 6174 is somewhat arbitrarily chosen. 
- Not sure if it's possible to make the port number a variable instead (so as to adhere to DRY principle); currently, the port number is repeated across multiple files)

Pros: macOS users don't need to switch off AirPlay. (Useful when, say, trying to use Mac to stream audio/video from iPhone!) 
Cons: Defies convention of using 5000 for ExpressJS. 6174 chosen [somewhat arbitrarily](https://en.wikipedia.org/wiki/6174_(number)).  It also appear [to be unassigned](https://www.adminsub.net/tcp-udp-port-finder/6174).

### DIFF
--- a/server/src/bootstrap/index.ts
+++ b/server/src/bootstrap/index.ts
@@ -77,7 +77,7 @@ app.use(requestLoggingMiddleware)
 
 // cors enable
 app.options('*', cors<express.Request>())
-app.use(cors({ origin: 'http://localhost:5000' }))
+app.use(cors({ origin: `http://localhost:${process.env.SERVER_PORT}` }))
 
 // security config
 app.use(helmet(helmetOptions))
--- a/.env.example
+++ b/.env.example
@@ -2,7 +2,7 @@ DB_HOST=localhost
 DB_NAME=askgov
 DB_USER=root
 DB_PASSWORD=password
-SERVER_PORT=5000
+SERVER_PORT=6174
 NODE_ENV=development
 SESSION_SECRET=mySessionSecret
 
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ Optionally [DBeaver](https://dbeaver.io/download/) to view database with GUI
   
   Frontend server accessible on `localhost:3000`
   
-  Backend server accessible on `localhost:5000/api/v1`
+  Backend server accessible on `localhost:6174/api/v1`
   
   Local mail server `MailDev` accessible on `localhost:1080`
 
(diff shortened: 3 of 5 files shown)


LoneRifle (changes requested):
Lgtm otherwise

server/src/bootstrap/index.ts
```diff
@@ -77,7 +77,7 @@ app.use(requestLoggingMiddleware)
 
 // cors enable
 app.options('*', cors<express.Request>())
-app.use(cors({ origin: 'http://localhost:5000' }))
+app.use(cors({ origin: 'http://localhost:6174' }))
```
LoneRifle: Have this replaced by `SERVER_PORT` instead

zxt-tzx: OK done!

LoneRifle (approved):
lgtm

PR #1052, opened 2021-12-30
fix(optionsmenu): fix height of topic selection in side menu

christopherlim98:
## Problem

For longer topics, the text will overflow for the side menu.

## Solution

Change style to use min-height instead of fixed height.

## Before & After Screenshots
**BEFORE**:
Screenshot 2021-12-30 at 11 07 32 AM

**AFTER**:
image

## Tests

Check that height of topic selection displays properly for longer topics (beyond 2 lines)


LoneRifle (approved):
lgtm!

PR #1048, opened 2021-12-29
feat(search-results): add breadcrumbs, update search query and search results UI

wanlingt:
## Problem

The search result page has not been updated according to the new designs on [Figma](https://www.figma.com/file/O6lxjjfn04Ow6nn2TMStpk/AskGov-Design-Master-v2?node-id=6255%3A56583)

Working towards #798 

## Solution

**Features**:

- Add breadcrumbs for navigation 
- Mention the search query in the headline
- Bold the search query terms whenever they occur in the search results' post title, ignoring case sensitivity
- Display answer preview, truncate if >120 characters

**Improvements**:
- Migrate theming from component to SearchResults Chakra theme

**Bug Fixes**:
- Make width of search results consistent for results found and no results found

## Before & After Screenshots

**BEFORE**:
Mobile:

https://user-images.githubusercontent.com/56983748/147716978-f4d125dd-623f-48bf-9daf-b74127186532.mov

Desktop:

https://user-images.githubusercontent.com/56983748/147716971-720acba7-14a1-4f63-81e3-451af4b681a3.mov

**AFTER**:
Mobile:

https://user-images.githubusercontent.com/56983748/147716806-e39ac187-525c-4c3d-bf27-5185e1379bf7.mov

Desktop:

https://user-images.githubusercontent.com/56983748/147716807-d5523e10-ce27-4eca-ba1f-d382d135fb0f.mov

## Tests

- Clicking on the breadcrumb links should bring you to the relevant pages
- Search terms should be bolded in the results
- Both question and answer should be displayed, clicking on either brings you to the post

PR #1036, opened 2021-12-28
refactor(optionsmenu/homepage/searchresults): add side menu for desktop layout v2

christopherlim98:
## Problem

Desktop layout is currently consistent with tablet layout. This does not make full use of screen real estate.

Closes #949 

## Solution

Updated desktop layout for home page and search results page.

- Create options side menu component
- Add listener on home page and search results page
- Add id to html elements for easier debugging
- Use componentMultiStyleConfig for `OptionsMenu` styles because DRY (e.g. `styles.accordionItem`)

**Potential Room for Improvements**:

- Currently there are 3 listeners to detect device type across `Header`, `SearchResults` and `HomePage` components. Would be good to explore a way to unify these 3 listeners by creating a shared context for device type. (#1045)
- To use id/classNames to label important components for easier debugging. 

## Screenshots
**HomePage: agency**

Screenshot 2021-12-28 at 7 42 51 PM

**HomePage: agency > topic**

Screenshot 2021-12-28 at 7 43 42 PM

**HomePage: agency > view all questions**

Screenshot 2021-12-28 at 7 45 52 PM

**Search**:
Screenshot 2021-12-28 at 7 45 09 PM

**Sticky Side Menu**:

https://user-images.githubusercontent.com/41635847/147563772-edb95d0f-6382-48f9-b89d-c6eefa4ed0f6.mov

## Tests

Check that:

1. desktop view is working according to figma design
2. mobile/tablet views are not broken due to changes

## Deploy Notes

Run locally using `npm run dev`

PR #1033, opened 2021-12-28
fix(homepage): display relevant agency's questions when logged in

wanlingt:
## Problem

When logged in, a user's agency's questions appear in another agency's page.  

Closes #781 

## Solution

Each agency's page lists questions from that agency. A user can only edit questions belonging to the user's agency.

**Bug Fixes**:

- `QuestionsListComponent` checks if the user's agency is the same as the agency of the page being shown, in addition to checking if there is an authenticated user

## Before & After Screenshots

**BEFORE**:

https://user-images.githubusercontent.com/56983748/147542743-5d26fb3f-b8cf-4d37-aa6d-7fc1fda35470.mov

**AFTER**:

https://user-images.githubusercontent.com/56983748/147542822-8e424ef7-e36a-4fa6-b3ba-4fbe1b3ed9cc.mov

## Tests

- Log in. The user's associated agency page should display posts which are editable. Other agencies' pages should show their agency's posts, and these should not be editable by the user.

PR #1062, opened 2022-01-03
Release 2021-01-03

LoneRifle:
This release contains the following major changes:
- #1033
- #1036 
- #1048
- #1052


wanlingt (approved):
tested, lgtm!

minor nitpick - the edit button for logged in users is out of place. In any case it still works + there are new designs for the edit and delete buttons, so we can leave it for the implementation of the new designs
image

christopherlim98:
> tested, lgtm!
> 
> minor nitpick - the edit button for logged in users is out of place (or was it already like this previously?). In any case it still works + there are new designs for the edit and delete buttons, so we can leave it for the implementation of the new designs image

Good catch, the 'delete' dropdown now occludes the bottom post.

client/src/theme/components/PostItem.ts
```diff
@@ -6,6 +6,7 @@ export const PostItem = makeMultiStyleConfig({
     py: '24px',
     borderBottom: '1px solid',
     borderBottomColor: 'neutral.300',
+    flexDirection: 'column',
```
christopherlim98, line 9: I think this causes the edit button to go below.

PR #1091, opened 2022-01-06
feat(search): highlight search query terms in search results

### LINKED ISSUE #1069: Highlight query matches in search results
All search terms are being highlighted, including stop words

image

Improve the relevance of highlighted keywords: Get OpenSearch to highlight query matches, then parse the results from the highlight object ([link](https://opensearch.org/docs/latest/opensearch/ux/#highlight-query-matches))

**Describe alternatives you've considered**
A less accurate alternative (given that we opted for fuzzy matching) would be to exclude stop words from being highlighted

wanlingt:
## Problem

The bolded search terms might include words that have not been used to identify the results e.g. stop words

Closes #1069 

## Solution

Add the `highlight` parameter to the query block, this returns a `highlight` object that shows the search term wrapped in a `<b></b>` tag. The highlight parameter highlights the original terms even when using synonyms or stemming for the search itself.

Relevant documentation [here](https://opensearch.org/docs/latest/opensearch/ux/#highlight-query-matches) and [here](https://docs.aws.amazon.com/opensearch-service/latest/developerguide/searching.html#searching-dsl)

## Before & After Screenshots

**BEFORE**:
Screenshot 2022-01-06 at 10 38 58 AM

**AFTER**:
image

## Tests

- Use misspelled search terms in the search query, the results shown should have the correct search term bolded

### DIFF
--- a/client/src/components/SearchBox/SearchBox.component.tsx
+++ b/client/src/components/SearchBox/SearchBox.component.tsx
@@ -17,7 +17,7 @@ import { RefCallBack } from 'react-hook-form'
 import { BiSearch } from 'react-icons/bi'
 import { useQuery } from 'react-query'
 import { Link, useNavigate } from 'react-router-dom'
-import { SearchEntry } from '~shared/types/api'
+import { SearchEntryWithHighlight } from '~shared/types/api'
 import { useGoogleAnalytics } from '../../contexts/googleAnalytics'
 import {
   getAgencyById,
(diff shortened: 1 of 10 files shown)


client/src/components/SearchBox/SearchBox.component.tsx
```diff
@@ -126,12 +126,12 @@ export const SearchBox = ({
 
   // null type required due to Downshift's type definitions
   // eslint-disable-next-line
-  const itemToString = (item: SearchEntry | null) =>
-    item?.title ? item.title : ''
+  const itemToString = (item: SearchEntryWithHighlight | null) =>
+    item?.result.title ? item.result.title : ''
```
LoneRifle:
```suggestion
    item?.result.title ?? ''
```

server/src/modules/search/__tests__/search.controller.spec.ts
```diff
@@ -50,30 +50,34 @@ describe('SearchController', () => {
   }
 
   describe('searchPosts', () => {
-    const searchEntries: SearchEntry[] = [
+    const searchEntries: SearchEntryWithHighlight[] = [
```
LoneRifle, line 53: do the tests verify if the highlight is present?

LoneRifle (changes requested):
lgtm otherwise

LoneRifle (approved):
lgtm =)

PR #1135, opened 2022-01-13
fix(secrets-check.sh): amend regex of AWS Key

### LINKED ISSUE #1134: Fix secrets-check script
I was looking into the `secrets-check.sh` yesterday and there is an error with the regex pattern. 
- [This](https://aws.amazon.com/blogs/security/a-safer-way-to-distribute-aws-credentials-to-ec2/) appears to be the canonical post from which these patterns originated
- I googled a little and it seems like (1) GoGovSG has the [same error](https://github.com/opengovsg/GoGovSG/blob/develop/secrets-check.sh) and (2) this erroneous pattern had been previously picked up and rectified in [this gist](https://gist.github.com/dduvnjak/ce08f917f7ead5f126ef)

I suspect it's probably an error made by an initial developer that propagated via copy-and-paste. It's quite a small issue and I intend to create a one-liner commit to fix this. (Should I do the same for GoGovSG?)

I researched a little around this issue and found that there are more involved solutions to this problem like [this](https://github.com/Yelp/detect-secrets) and [this](https://github.com/awslabs/git-secrets) (my cursory impression is the latter seems like a better fit, in case we ever want to go ahead; [this](https://seandavi.github.io/post/protect-against-secrets-in-git-repositories/) seems to be a war story/tutorial of sorts).

zxt-tzx:
Previous AWS key regex was wrong. It is now correct.

fixes #1134

### DIFF
--- a/secrets-check.sh
+++ b/secrets-check.sh
@@ -7,7 +7,7 @@ exec 1>&2
 
 # Check changed files for an AWS keys
 KEY_ID=$(git diff --cached --name-only -z $against | xargs -0 cat | perl -nle'print $& if m{(?<![A-Z0-9])[A-Z0-9]{20}(?![A-Z0-9])}')
-KEY=$(git diff --cached --name-only -z $against | xargs -0 cat | perl -nle'print $& if m{(?<![A-Z0-9])[A-Z0-9]{20}(?![A-Z0-9])}')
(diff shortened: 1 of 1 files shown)


LoneRifle (approved):
lgtm

PR #1196, opened 2022-01-25
Fix typo in previous regex

zxt-tzx:


christopherlim98 (approved):
lgtm, no longer seeing the false positive

PR #1347, opened 2022-02-25
Refactor HomePage component in order to add topics management functionality

zxt-tzx:
## Problem

I am trying to refactor the HomePage component as a precursor to adding the topics management functionality. 

## Solution

I have split HomePage into HomePage (for AskGov as a whole) and AgencyHomePage. I have clarified with @khaleedah that the current coupling is unnecessary and (long-term) undesirable.

I have extracted out the components common to both HomePage and AgencyHomePage; the hierarchy of components is now: 
- Questions 
  - QuestionsHeader
    - Text
    - SortQuestionsMenu
    - PostQuestionButton
  - QuestionsList (subcomponents therein unmodified)

I have also added comments in AgencyHomePage component discussing next steps to refactor and/or build out the topics management functionality. These will be removed as the functionality is built out.

## Questions

1. Should I continue to refactor the components in AgencyHomePage that are not used in HomePage? (@prestonlimlianjie had also previously advised that the components themselves should render themselves based on the viewport size, rather than handling them at the page level.) One the one hand, this is good for code cleanliness + if we want to split up AgencyHomePage into a separate TopicsPage (in anticipation of nested topics in the future). On the other hand, it's orthogonal to the task of building out topics management functionality.  (See https://github.com/opengovsg/askgovsg/pull/1350 for more discussion.) 

2. Based on the current Figma design (https://www.figma.com/file/O6lxjjfn04Ow6nn2TMStpk/AskGov-Design-Master-v2?node-id=6547%3A87329), it seems like the topics management functionality is only envisioned on the AgencyHomePage (where the topics are shown). Currently, the same component `OptionsMenu` is used on both the general HomePage (to display the list of agencies) and on the AgencyHomePage (to display the list of topics). This same component is then used in a third view (after clicking into a topic + the viewport size is non-desktop). 

As such, I propose to split the `OptionsMenu` into three different subcomponents:
- AgenciesMenu, to be used on HomePage
- TopicsManagementMenu, to be used on AgencyHomePage with topics management functionality
- TopicsSelectorMenu, to be used in the last topic + non-desktop viewport view


zxt-tzx:
I don't really understand what this line does
image

Would this work just as well? (The non-null assertion operator notwithstanding)
image

Or maybe something like this (not sure how using `as` is different from non-null assertion operator, except insofar as it doesn't trip up ESLint hmm)
image

More generally, if `agency` is undefined, then what exactly are the topics being fetched? (Some random topics?) I see that this runs in the AskGov home page, but it's never displayed anywhere afaict

(Is this pre-fetching? Gonna go do some tutorials on `react-query` hmm)

prestonlimlianjie (approved):
Lgtm overall. Haven't tested it locally because of setup issues/lack of time, but I've taken a look through the code. Let's chat more in our 1-1!

A few recommendations that we could tackle in separate issues and PRs, pending product priorities:

### Explore using React Context
We could consider using [React Context](https://reactjs.org/docs/context.html) to avoid prop drilling across
- setSortState, sortState, questionDisplayState: **Questions -> QuestionsHeader -> SortQuestionsMenu**
- setQuestionDisplayState: **Questions -> QuestionsList**

### Create primitive components instead of using Chakra UI components directly
One of the challenges that I notice from the code base is the length of the higher-level component files. Part of the problem is that we are using Chakra UI components directly, and as such, we need to specify lots of granular config for each component. E.g.
```
          <MenuButton
            as={Button}
            variant="outline"
            borderColor="secondary.700"
            color="secondary.700"
            borderRadius="4px"
            borderWidth="1px"
            w={{ base: '100%', sm: '171px' }}
            textStyle="body-1"
            textAlign="left"
            d={{
              base: questionsDisplayState.value === 'all' ? 'block' : 'none',
            }}
          >
```

A way to overcome this is to create another abstraction layer above Chakra UI, for example, we could come up with an AskGov-specific MenuButton that already captures the relevant defaults for the AskGov use case. In that way, we could reduce the number of params we pass into the MenuButton component. Imagine if all we had to write was the following:
```
          <MenuButton
            variant="primary"
            d={{
              base: questionsDisplayState.value === 'all' ? 'block' : 'none',
            }}
          >
```

The component files will be less daunting to review and understand.

zxt-tzx:
I notice there are some unintended changes in functionalities with this refactoring. Specifically, instead of showing `QUESTIONS ON THIS TOPIC` after clicking into the topic, it shows `TOP QUESTIONS` 😢 

I think to fix this elegantly, I'll have to use React Context. I'll continue to refactor this branch and re-request review when I'm done!

UPDATE: have refactored this. Code is cleaner now, but the coupling of topics and questionDisplayState is still bad (weirdness hidden away in `Questions.component.tsx` for now). 

Also not sure why this check is failing:
image

zxt-tzx:
@christopherlim98 could I trouble you to look at the changes I made to `QuestionsList.stories.tsx` before I merge it in? also, considering I added various components, wonder whether we need to add `*.stories.tsx` files for those too. 

I must say I still don't fully understand what Storybook does!

Thanks!

christopherlim98:
> @christopherlim98 could I trouble you to look at the changes I made to `QuestionsList.stories.tsx` before I merge it in? also, considering I added various components, wonder whether we need to add `*.stories.tsx` files for those too.
> 
> I must say I still don't fully understand what Storybook does!
> 
> Thanks!

Hey! Storybook is meant to be a way to do UI documentation, and to pave the way for Chromatic integration to allow UI testing.

In your local dev, you can run `npm run storybook` which will launch the storybook console and allow you to see these components in isolation. 

Currently, a few of the stories fail to load (e.g.AgencyHomePage, OptionsSideMenu, ...) With these new changes, you should update the storybook `preview.tsx` file to include the newly created `HomePageProvider` so that the storybook components will work as expected. 

After discussion with Khaleedah, Storybook does not need to be so granular to show every component. Will PM you the storybook working document so that you can see what kind of items will go over there. 

```
const withChakra = (StoryFn: Function) => {
  return (
    <ChakraProvider theme={theme}>
      <MemoryRouter>
        <QueryClientProvider client={queryClient}>
          <AuthProvider>
            <HelmetProvider>
              <HomePageProvider>
                <StoryFn />
              </HomePageProvider>
            </HelmetProvider>
          </AuthProvider>
        </QueryClientProvider>
      </MemoryRouter>
    </ChakraProvider>
  )
}
```

christopherlim98 (changes requested):
See comment above^

client/src/contexts/HomePageContext.tsx
```diff
@@ -0,0 +1,52 @@
+import { createContext, ReactChild, ReactChildren, useState } from 'react'
+import {
+  DEFAULT_QUESTIONS_DISPLAY_STATE,
+  DEFAULT_QUESTIONS_SORT_ORDER,
+  QuestionsDisplayState,
+  QuestionsSortOrder,
+} from '../components/Questions/questions'
+
+interface HomePageProps {
+  children: ReactChild | ReactChild[] | ReactChildren | ReactChildren[]
```
christopherlim98, line 10: Was wondering if this can be replaced with JSX.Element?

zxt-tzx: oh yeah I replaced with `ReactNode` in the other AgencyHomePage PR

zxt-tzx:
@christopherlim98 pls help me check if this is OK :) https://github.com/opengovsg/askgovsg/pull/1347/commits/a2e1343f66fe476520b2910f8e0db6880e882a56

christopherlim98 (approved):
The SearchResult story seems to be broken. But I think we can go ahead with this PR and just open a new issue to fix that story at a later date.

PR #1349, opened 2022-02-25
Add eslint rules on import order and apply to src/client

zxt-tzx:
Slightly opinionated but I really like the import order rules from `ts-template` and have set it up + applied it to the `src/client`

### DIFF
--- a/client/.eslintrc.js
+++ b/client/.eslintrc.js
@@ -3,10 +3,49 @@ module.exports = {
     'plugin:@typescript-eslint/recommended',
     'plugin:prettier/recommended',
   ],
-  plugins: ['@typescript-eslint'],
+  plugins: ['@typescript-eslint', 'import', 'simple-import-sort'],
   rules: {
     'no-console': 'warn',
     'no-undef': 'error',
+    // Rules for auto sort of imports
+    'simple-import-sort/imports': [
+      'error',
+      {
+        groups: [
+          // Side effect imports.
+          ['^\\u0000'],
+          // Packages.
+          // Packages. `react` related packages come first.
+          // Things that start with a letter (or digit or underscore), or
+          // `@` followed by a letter.
+          ['^react', '^@?\\w'],
+          // Root imports
+          // Shared imports should be separate from application imports.
+          ['^(~shared)(/.*|$)'],
+          ['^(~)(/.*|$)'],
+          ['^(~typings)(/.*|$)'],
+          [
+            '^(~assets|~theme)(/.*|$)',
+            '^(~contexts)(/.*|$)',
+            '^(~constants)(/.*|$)',
+            '^(~hooks)(/.*|$)',
+            '^(~utils)(/.*|$)',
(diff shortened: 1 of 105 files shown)


client/.eslintrc.js
```diff
@@ -3,10 +3,49 @@ module.exports = {
...
+          // Packages. `react` related packages come first.
+          // Things that start with a letter (or digit or underscore), or
+          // `@` followed by a letter.
+          ['^react', '^@?\\w'],
+          // Root imports
+          // Shared imports should be separate from application imports.
+          ['^(~shared)(/.*|$)'],
+          ['^(~)(/.*|$)'],
+          ['^(~typings)(/.*|$)'],
+          [
```
christopherlim98: Question: how are these rules generated? I think the config for lines 27-37 may not be applied for our repo as we do not use alias imports of the structure e.g. `~assets` in this repository.

christopherlim98 (commented):
LGTM otherwise.

While we are trying to fix the imports, thoughts on using alias imports instead of relative imports? Similar to https://github.com/opengovsg/petitionsgovsg/pull/291

client/.eslintrc.js
zxt-tzx: haha I straight up copied directly from `ts-template`!

I did see these kinda imports in the codebase though (is this what you meant by alias imports?):
image

christopherlim98: Yes, this repo uses an alias for shared imports (can checkout `tsconfig.paths.json`), which is covered by line 24 of this config. But I think the other lines of config added here from 27-37 will not be used since we don't use alias imports for those folders in this repo.

christopherlim98: To check out why ts-template uses that particular setting, that is because they use alias imports for those folders here https://github.com/opengovsg/ts-template/blob/develop/frontend/tsconfig.paths.json

zxt-tzx: Ah I see your point. OK will remove lines 27-37.

zxt-tzx:
Hmm personally I haven't had much trouble with relative imports in practice because I just click through to the original file and the IDE will show the folder structure leading to that file. 

For https://github.com/opengovsg/petitionsgovsg/pull/291, did you have to manually go in and swap up all the relative imports with alias imports? Wonder if there's a library or sth that can automate that, if we do set that up.

client/.eslintrc.js
```diff
@@ -24,17 +24,6 @@ module.exports = {
           ['^(~shared)(/.*|$)'],
           ['^(~)(/.*|$)'],
           ['^(~typings)(/.*|$)'],
```
christopherlim98: nit: i think line 25 and 26 can be removed too since they are unused for now

christopherlim98 (approved):
LGTM! Thanks for cleaning up the imports 🙇

client/.eslintrc.js
zxt-tzx: OK removed!

PR #1350, opened 2022-02-25
Build out topics management functionality with dedicated TopicsManagementMenu

### LINKED ISSUE #747: Topics management for agencies
Allow agencies to create / edit / delete topics on agency dashboard

Figma:
https://www.figma.com/file/O6lxjjfn04Ow6nn2TMStpk/AskGov-Design-Master-v2?node-id=6547%3A87329

zxt-tzx:
Closes https://github.com/opengovsg/askgovsg/issues/747

NB: to rebase on develop after https://github.com/opengovsg/askgovsg/pull/1347 and https://github.com/opengovsg/askgovsg/pull/1383 have been merged in:
- Clean up `TopicsMenu.component.tsx` (use React Context, remove comments etc.)
- Put `TopicsMenu` in `homePageDefaultView` (not needed in `homePageAllQuestionsView`? To test on staging: what happens when you log in + click view all questions OR vice versa)

---
Creating TopicsManagementMenu based on existing OptionsMenu (which itself will be split). 

Checked with @khaleedah, currently the plan is to only support one layer of topics.

As noted in the comments, to make this component truly extensible (i.e. to support nested topics), it should (1) detect whether it exists at an AgencyHomePage or within a topic and (2) render the level of topics accordingly. The interface should be consistent across both instances (which is not the case now; when clicked into a topic, the menu is folded).

Not super sure how to build this; have gone with a simpler one layer topics. 

Also, current interface does not provide for adding topics description.

See known UI bug in comment below.


zxt-tzx:
Have built out out topics management functionality. 

UI bug with `AddNewTopicCard` that I am not sure how to solve; to repro:
1. Swap out `OptionsMenu` with `TopicsMenu` in `HomePage.component.tsx` (this is for testing purposes only)
2. Log in as agency user
3. Go to `/agency/${agencyShortName}`
4. Click on "Add new topic" and enter arbitrary topic name
5. Click on the "tick" button (NB this is critical; pressing Enter doesn't result in buggy behavior)
image

Issue: `Editable` remains in `isEditing` state
image

Not sure how to fix, have tried all solutions I can think of 😢 Not sure if it's a Chakra UI problem

***

UPDATE: fixed here https://github.com/opengovsg/askgovsg/pull/1350/commits/ee9126ab1046bd3036302a8e604fd78efc3d6430

Takeaway: to be more careful about clicks propagating to parent components.

zxt-tzx:
h/t to @orbitalsqwib for https://github.com/opengovsg/askgovsg/pull/1350/commits/c9b4cc1c15d2dd6b17adebcee776bb8ba6369d40

zxt-tzx:
Heads up that this PR is ready for review :) @LoneRifle

LoneRifle (changes requested):
lgtm otherwise. You have to do a bugfix to cater for the scenario where the user tries to delete a topic where there are still questions under it. Right now, the questions are orphaned, and attempting to edit the orphaned questions crashes the UI since it tries to access `topic.id` even though topic is undefined.

There are a couple of ways you can tackle this:
- Prevent the user from deleting a topic if we know there are questions under it
- Tweak the frontend so that orphaned questions are still editable.

zxt-tzx:
Ohh that sounds like a problem with the delete API? According to the Figma/current user prompt, the plan is to warn the user that all questions under it should be deleted as well. If we go with that, I can open a separate PR to modify the delete API?

image

I'm not sure which is a better/worse:
- User who wants to delete topics and all its questions must delete the questions individually first (?)
- User who only meant to delete a topic but not the questions accidentally deletes all of them (despite our prompt) 
 
[As an aside, I noticed that the delete API is weird in that it **actually deletes** the rows from the database, rather than soft deleting the row, which would suggest it's irreversible -> not sure why it's designed this way hmm I don't think the memory saved is significant...]

LoneRifle:
I think for safety reasons, let's just go with preventing the user performing the delete if there are questions under the topic, since we should leave it to the user to decide what to do with the questions (delete, move under new topic, or etc). Hard deletes were done over soft deletes given that they were simpler to reason about, especially given its relatively early stage of development at the time.

zxt-tzx:
@LoneRifle have implemented the feature 

SCR-20220407-t1e

SCR-20220407-t1b

LoneRifle (approved):
lgtm

PR #1383, opened 2022-03-05
Refactor AgencyHomePage to make four possible views explicit

zxt-tzx:
## Problem

To pave the way for easy incorporation of topics management functionality in https://github.com/opengovsg/askgovsg/pull/1350 

Extends https://github.com/opengovsg/askgovsg/pull/1347 and should be merged in after https://github.com/opengovsg/askgovsg/pull/1347 is merged. 

## Solution

Refactor AgencyHomePage to make four possible views explicit

### DIFF
--- a/client/src/components/Questions/Questions.component.tsx
+++ b/client/src/components/Questions/Questions.component.tsx
@@ -72,7 +72,7 @@ export const Questions = ({
           questionsPerPage || questionsDisplayState.questionsPerPage
         }
         showViewAllQuestionsButton={
-          showViewAllQuestionsButton ?? questionsDisplayState.value !== 'all'
+          showViewAllQuestionsButton ?? questionsDisplayState.value === 'top'
         }
         listAnswerable={listAnswerable}
       />
--- a/client/src/components/OptionsMenu/OptionsMenu.component.tsx
+++ b/client/src/components/OptionsMenu/OptionsMenu.component.tsx
@@ -26,8 +26,6 @@ import {
   LIST_AGENCY_SHORTNAMES,
 } from '../../services/AgencyService'
 import {
-  fetchTopics,
-  FETCH_TOPICS_QUERY_KEY,
   getTopicsUsedByAgency,
   GET_TOPICS_USED_BY_AGENCY_QUERY_KEY,
 } from '../../services/TopicService'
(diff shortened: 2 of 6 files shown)


christopherlim98:
Could you change the base branch to #1347  instead of develop so that the diff is easier to read? :)

LoneRifle:
> Could you change the base branch to #1347 instead of develop so that the diff is easier to read? :)

ftfy

LoneRifle (changes requested):
you need to provide more information in the commit message and/or PR to describe _how_ the refactor was done, as it's not easily inferred from the diff. See https://cbea.ms/git-commit/ for more information.

client/src/components/Questions/Questions.component.tsx
```diff
@@ -72,7 +72,7 @@ export const Questions = ({
           questionsPerPage || questionsDisplayState.questionsPerPage
         }
         showViewAllQuestionsButton={
-          showViewAllQuestionsButton ?? questionsDisplayState.value !== 'all'
+          showViewAllQuestionsButton ?? questionsDisplayState.value === 'top'
```
LoneRifle, line 75: Also, why did you change this?

zxt-tzx: ah this was in response to a bug I found. `showViewAllQuestionsButton` is not a great name (but I was struggling to figure out what a better name is) because what it does is it either shows the "View all questions" button or it shows the pagination. 

After checking out the live AskGov, I noticed the behavior we want is to show the "View all questions" when viewing top questions. 

In the two other display states (i.e. currently viewing all questions OR viewing the questions of a selected topic), we really want to show the pagination. 

Hence this change!

LoneRifle: are you sure about the latter?

zxt-tzx: I... think so? I guess to put it more succinctly, the "View all questions" button should only show up when:
- User is looking at the top questions, so the button brings them to all questions instead
- User is authenticated, in which case they'll be straight-up shown 50 questions, with no need to toggle to see more (I'm assuming the original design intent is 50 questions should be more than enough to see all questions...? I'm guessing this hard-coding might result in errors down the line...?)

In most cases, I don't think it makes a difference because users will typically access topic via the AgencyHomePage, but changing this ensures the following are consistent:

Local without this change:
image

Live AskGov:
image

LoneRifle (approved):
lgtm, merge when ready

PR #1389, opened 2022-03-07
build(deps-dev): remove react-tabs, not used

LoneRifle:
react-tabs was used in earlier iterations of AskGov before Chakra was adopted as component framework of choice. Remove react-tabs accordingly


christopherlim98 (approved):
lgtm!

PR #1443, opened 2022-03-23
feat(OptionsMenu): display agency long and short names

nigel0511:
Fixes #1405 
added features:

1. useQuery to request list of all agency data
2. map agency data to display both long and short name

PR #1427, opened 2022-03-18
backend "list all agencies" route, controller

nigel0511:
#1405 
I have implemented a route to request list of all agencies from database, please comment.

PR #1444, opened 2022-03-23
Release 2022-03-23

LoneRifle:
This release contains the following major changes:

- #1347 
- #1383 
- #1443 
  - #1427 
  - #1433


christopherlim98 (approved):
lgtm!
minor nit: the style for accordion menu items break if they are > 2 lines. However, I think the names in production are not that long. We can patch it in a future update.

PR #1449, opened 2022-03-24
style(Agency/OptionsMenu): increase OptionsMenu max-width to 680px and fix styles for AgencyHomePage

christopherlim98:
Changes are live on staging.

## Problem

- Options menu is set to 620px max width, when it should be 680px. This causes option items to wrap into 3 lines when it should be 2 lines.
- Tablet view ( < 1440px) would cause the topic description and question header to fold into one single row.
- Mobile view shows that the topic description has a large bottom margin.
- `getRedirectUrlTopics` should not accept optional agency shortname.

## Solution
- Set options menu max width to 680px.
- Increase breakpoint for AgencyHomePage from l (1024px) to xl (1440px)
- Use responsive bottom margin for topic description.
- Refactor `getRedirectUrlTopics` to take in a string field for agencyShortName.

## Before & After Screenshots

**BEFORE**:
Tablet view ( < 1440px) would cause the topic description and question header to fold into one single row.
Screenshot 2022-03-24 at 3 13 03 PM

Mobile view shows that the topic description has a large bottom margin.
Screenshot 2022-03-24 at 3 13 35 PM

**AFTER**:
Screenshot 2022-03-24 at 3 22 50 PM

Screenshot 2022-03-24 at 3 22 35 PM

### DIFF
--- a/client/src/components/OptionsMenu/OptionsMenu.component.tsx
+++ b/client/src/components/OptionsMenu/OptionsMenu.component.tsx
@@ -93,6 +93,18 @@ const OptionsMenu = (): ReactElement => {
     }
   })
 
+  const OptionsItem = (option: string) => {
+    return (
+      <Flex m="auto" w="100%" px={8}>
+        <Text maxW="244px">{option}</Text>
+        <Spacer />
+        <Flex alignItems="center">
+          <BiRightArrowAlt />
+        </Flex>
+      </Flex>
+    )
+  }
+
   const optionsMenu = (
     <SimpleGrid sx={styles.accordionGrid}>
       {agency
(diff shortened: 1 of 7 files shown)


LoneRifle (approved):
lgtm!

christopherlim98:
> lgtm!

Will merge after checking once more with Khaleedah

christopherlim98:
Changes since last commit:
- Previously the y-axis paddings were controlled by both `HomePage` and `AgencyPage`. Decided to shift the y-axis padding over to `QuestionHeader`. 
- Increase width to 680px for both `HomePage` and `AgencyPage` for the tablet and desktop views.

PR #1448, opened 2022-03-24
refractor: useDetectDevice hook

nigel0511:
1. extract detect device function from Header, AgencyHomePage, SearchResults components
2. create new useDetectDevice hook under hooks file
3. create getCurrentDeviceSize() for duplicated code

PR #1452, opened 2022-03-26
refactor: useDetectDevice hook

### LINKED ISSUE #1045: Use common listener for device type
Currently there are 3 listeners to detect device type across Header, SearchResults and HomePage components. 

Would be good to explore a way to unify these 3 listeners by creating a shared context for device type.

nigel0511:
Fixes #1045 

replace pull request #1448 

1. extract detect device functions from Header, SearchResults and AgencyHomePage component.
2. create useDetectDevice hook in hooks file and create getCurrentDeviceSize() for replicated code.
3. replace detect device functions with useDetectDevice.

### DIFF
--- a/client/src/hooks/useDetectDevice.ts
+++ b/client/src/hooks/useDetectDevice.ts
@@ -1,12 +1,7 @@
 import { useState, useEffect } from 'react'
+import { device } from '../util/devicetype'
 
 export const useDetectDevice = () => {
-  const device = {
-    mobile: 'mobile',
-    tablet: 'tablet',
-    desktop: 'desktop',
-  }
-
   const getCurrentDeviceSize = () => {
     return window.innerWidth < 480
       ? device.mobile
--- a/client/src/hooks/useDetectDevice.ts
+++ b/client/src/hooks/useDetectDevice.ts
@@ -1,13 +1,18 @@
 import { useState, useEffect } from 'react'
-import { device } from '../util/devicetype'
+
+export enum DeviceType {
+  Mobile = 'mobile',
+  Tablet = 'tablet',
+  Desktop = 'desktop',
+}
 
 export const useDetectDevice = () => {
   const getCurrentDeviceSize = () => {
     return window.innerWidth < 480
-      ? device.mobile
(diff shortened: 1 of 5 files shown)


client/src/hooks/useDetectDevice.ts
```diff
@@ -0,0 +1,30 @@
+import { useState, useEffect } from 'react'
+
+export const useDetectDevice = () => {
+  const device = {
+    mobile: 'mobile',
+    tablet: 'tablet',
+    desktop: 'desktop',
+  }
```
LoneRifle: Consider making this an enum exported from this module. Perhaps something like..

```ts
export enum Device {
  Mobile = 'mobile',
  // ...
}

File: client/src/pages/AgencyHomePage/AgencyHomePage.component.tsx [hunk 2]
```diff
@@ -238,7 +212,7 @@ const AgencyHomePage = (): JSX.Element => {
         ? homePageDefaultView // no topics param key -> default homepage
         : !topicQueried
         ? homePageAllQuestionsView // topics param key + no topic queried -> all questions
-        : deviceType === device.desktop // specific topic selected
+        : deviceType === 'desktop' // specific topic selected
```
LoneRifle: we shouldn't be hardcoding string literals here, as we then lose the constraints that limit the comparison of `deviceType` to specific values. import the proposed enum mentioned elsewhere, and use those values instead.

LoneRifle (changes requested):
Have a look at the things I've highlighted

client/src/pages/AgencyHomePage/AgencyHomePage.component.tsx
nigel0511: okay, I will make the suggested change, thank you for pointing that out.

nigel0511:
Change made:
1. created devicetype.ts under util, export enum device.
2. replace string literals with variables 

please advise if that's appropriate, open to suggestions.

LoneRifle (approved):

PR #1521, opened 2022-04-08
build(deps-dev): reinstate wildcard for webpack

LoneRifle:
## Problem and Solution
webpack issues with Chakra seem to have resolved with webpack/webpack#15634, so reinstate wildcard


zxt-tzx:
updated the `package-lock.json`, otherwise lgtm

zxt-tzx (approved):

PR #1523, opened 2022-04-08
Add x button and bold important text in topics management dialog

### LINKED ISSUE #1522: [frontend] add x button to ConfirmDialog and FailureDialog
image

zxt-tzx:
Closes https://github.com/opengovsg/askgovsg/issues/1522

Some notes:
- Extra padding added to FailureDialog so as to maintain spacing between words and the CloseButton. This is probably not great for reusing and extending this component? 
- Bolding is slightly more difficult because if you pass tags like `<chakra.span>` as props, they will get rendered as literal strings. The way around that is to use `dangerouslySetInnerHTML` in conjunction with sanitizing HTML. Strictly speaking, bolding number of questions is quite safe because users can only control the number of questions, which is not dangerous. But sanitized anyway. For bolding the topic, sanitization is necessary, 

Results:
image
image

### DIFF
--- a/client/src/components/ConfirmDialog/ConfirmDialog.component.tsx
+++ b/client/src/components/ConfirmDialog/ConfirmDialog.component.tsx
@@ -11,6 +11,7 @@ import {
   CloseButton,
   HStack,
 } from '@chakra-ui/react'
+import sanitizeHtml from 'sanitize-html'
 
 interface ConfirmDialogProps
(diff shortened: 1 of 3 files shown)


LoneRifle (approved):
Something is up with CI - for _some_ reason npm is struggling to verify the integrity of one of the modules for `/client`. I'm going to squash and merge and hope for the best

PR #1537, opened 2022-04-12
fix(ci): use new npm cache key

LoneRifle:
## Problem and Solution
Our GitHub Actions cache for npm is corrupted,
breaking CI. Specify a different key prefix to start
a fresh set of npm dependencies to cache.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,10 +45,9 @@ jobs:
         with:
           # npm cache files are stored in `~/.npm` on Linux/macOS
           path: ~/.npm
-          key: ${{ runner.OS }}-node-${{ hashFiles('**/package-lock.json') }}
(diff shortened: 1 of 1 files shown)


christopherlim98:
~~I think we might need to add the cache key change to the Test workflow too.~~

Also, I was wondering if it is worth while to consider putting the [cache_version as a github secret](https://stackoverflow.com/a/64819132)? So that we do not need to make a workflow change each time this crops up?

E.g.
`${{ runner.OS }}-node-askgov-${{ secrets.CACHE_VERSION }}-${{ hashFiles('**/package-lock.json') }}`

LoneRifle:
> Also, I was wondering if it is worth while to consider putting the [cache_version as a github secret](https://stackoverflow.com/a/64819132)? So that we do not need to make a workflow change each time this crops up?

Not comfortable with the idea - I think it's worthwhile to explicitly document when we were forced to change the keys

christopherlim98 (approved):
lgtm!

PR #1549, opened 2022-04-12
feat(char-count): add character exceed logic

### LINKED ISSUE #1053: When user types in >150 character in question input, show a clearer error message
**Describe the bug**
As most agency users are copy pasting FAQ from their mastersheet, the character count doesn't count down but instead just shows zero

**To Reproduce**
1. Login
2. Click on 'Post Question'
3. Paste into the question box 

image

**Expected behavior**
Suggestion: Provide clearer error messages

If user is over character limit, in real time, show `X characters over`
(see on right)
image

christopherlim98:
## Problem

Closes #1053

## Solution

* BorderFocusColor set to `error.500` when there is an error
* BorderFocusColor set to `secondary.700` when there is no error
* Do not just show `0 characters left` when characters exceeded
* Add icon next to character count warning when characters exceed 150

## Before & After Screenshots
**BEFORE**:
Screenshot 2022-04-12 at 7 05 39 PM

**AFTER**:
Screenshot 2022-04-13 at 10 17 37 AM

## Contextual Details
Update: after discussing with Khaleedah, we have decided to remove the error icon from the input group.
Screenshot 2022-04-12 at 7 07 54 PM

### DIFF
--- a/client/src/pages/PostForm/AskForm/AskForm.component.tsx
+++ b/client/src/pages/PostForm/AskForm/AskForm.component.tsx
@@ -15,8 +15,6 @@ import {
   HStack,
   Icon,
   Input,
-  InputGroup,
-  InputRightElement,
   useMultiStyleConfig,
 } from '@chakra-ui/react'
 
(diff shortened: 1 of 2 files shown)


LoneRifle (approved):
lgtm!

PR #1554, opened 2022-04-13
feat(rte): update style and add toolbar options 

### LINKED ISSUE #815: RTE field expands based on users input length
On the current RTE, agency users have to scroll within the input field itself.

This is not ideal as they need to do a lot of scrolling to check longer form answers / after inserting images which takes up a lot of real estate.

Currently am thinking that we can expand the answer field RTE as user types (just like Postman), but open to other suggestions as well.

christopherlim98:
## Problem
Closes #815

The rich text editor is missing the following features since the latest design update:
- Toolbar options: `strikethrough`, `textalign`
- Styling: grey header with height of 48px
- Styling: to use boxicons [Not fixed in this PR]
- Styling: for RTE textarea to expand when user input overflows

## Solution
This PR aims to solve the first two issues by adding the `strikethrough` and `textalign` toolbar options, and updating the styles. 

**This PR does not convert the RTE default icons to BoxIcons since the changes required are rather extensive.** By default, the RTE library only accepts icons in the format of a `path`, `url` or `base64encoded image` (see [issue](https://github.com/jpuri/react-draft-wysiwyg/issues/872)). Unfortunately, we are not able to just pass in the React Component form `react-icons/bi` out-of-the-box. This change can be deferred for a later PR. 

**Context**:
Screenshot 2022-04-13 at 4 28 06 PM
Screenshot 2022-04-14 at 4 59 28 PM

## Before & After Screenshots

**BEFORE**:
Screenshot 2022-04-13 at 4 29 16 PM

**AFTER**:
Screenshot 2022-04-13 at 4 29 23 PM

https://user-images.githubusercontent.com/41635847/163357817-52788170-c496-48fc-8016-554adb4a2cdd.mov

## Test
Changes should be live on staging in a bit.

### DIFF
--- a/client/src/components/RichText/ImageControl.tsx
+++ b/client/src/components/RichText/ImageControl.tsx
@@ -318,7 +318,7 @@ export const ImageControl = ({
   )
 
   return (
-    <Box display={{ base: 'none', xl: 'block' }}>
+    <Box display={{ base: 'none', xl: 'block' }} mb="6px">
       <Box onClick={onImageModalOpen} className="rdw-option-wrapper">
         <BiImage size="18px" />
       </Box>
--- a/client/src/components/RichText/LinkControl.tsx
+++ b/client/src/components/RichText/LinkControl.tsx
@@ -90,11 +90,11 @@ export const LinkControl = ({
     </div>
   )
   return (
-    <div className={styles.linkControl}>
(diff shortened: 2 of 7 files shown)


LoneRifle (approved):
lgtm, good work!

PR #1564, opened 2022-04-17
feat: add backend safeguard against deleting topic that has posts

zxt-tzx:
## Problem

Following https://github.com/opengovsg/askgovsg/pull/1350, I was somewhat uncomfortable with the fact that deleting a topic that will result in orphan posts (with no parent topic) is only safeguarded on the frontend. 

## Solution

Have modified the backend slightly to add an additional layer of safeguard

## Doubts

I'm not very familiar with Sequelize, so am not very sure if that part of `post.service.ts` is correct. Would appreciate if reviewer could take a closer look at that.

### DIFF
--- a/server/src/modules/topics/topics.controller.ts
+++ b/server/src/modules/topics/topics.controller.ts
@@ -175,9 +175,8 @@ export class TopicsController {
     }
 
     // Check if topic has posts
-    const hasPosts = await this.postService
-      .getPostsOfTopic(topicId)
-      .then((data) => data.totalItems !== 0)
+    const posts = await this.postService.getPostsOfTopic(topicId)
+    const hasPosts = posts.totalItems !== 0
 
     if (hasPosts) {
       return res
--- a/server/src/modules/topics/topics.controller.ts
+++ b/server/src/modules/topics/topics.controller.ts
@@ -7,20 +7,25 @@ import { Message } from '../../types/message-type'
 import { StatusCodes } from 'http-status-codes'
 import { ControllerHandler } from '../../types/response-handler'
 import { TopicWithChildRelations } from './topics.service'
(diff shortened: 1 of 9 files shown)


LoneRifle (changes requested):
- a better name for the func would be `getPostsByTopic`
- could you provide a test that exercises the case when `hasPosts` on topic deletion?

zxt-tzx:
done

server/src/modules/topics/topics.controller.ts
```diff
@@ -169,6 +174,16 @@ export class TopicsController {
         .json({ message: 'You do not have permission to delete this topic' })
     }
 
+    // Check if topic has posts
+    const posts = await this.postService.getPostsByTopic(topicId)
+    const hasPosts = posts.totalItems !== 0
+
+    if (hasPosts) {
+      return res
+        .status(StatusCodes.FORBIDDEN)
```
LoneRifle:
```suggestion
        .status(StatusCodes.BAD_REQUEST)
```
403 (and the related 401) deals with permission-related issues, not so much the correctness of a request

LoneRifle (changes requested):
Just one more. FIx this up and adjust the tests as needed

server/src/modules/topics/topics.controller.ts
zxt-tzx: OK

zxt-tzx: was debating which to use and tbh neither seemed like a perfect fit!
image
image

zxt-tzx: done

LoneRifle (approved):

PR #1553, opened 2022-04-13
style(header): add box shadow

christopherlim98:
## Problem

Current styling for the headers is not visually similar to Figma design.

## Solution

Add missing box shadow.

## Before & After Screenshots

**BEFORE**:
Screenshot 2022-04-13 at 11 12 41 AM
Screenshot 2022-04-13 at 11 13 11 AM

**AFTER**:
Screenshot 2022-04-13 at 11 07 08 AM
Screenshot 2022-04-13 at 11 07 39 AM

PR #1548, opened 2022-04-12
style(edit-question): update styles for post/edit/delete button

christopherlim98:
## Problem

There are a few style updates for the following buttons:
- `Edit` button
- `Delete` button
- `Post` button

There are also a few UI bugs that need to be fixed, such as incorrect margins between `Post` button and `AccordionMenu`
Closes #1517
Closes #1547

## Solution
Previously, we only have a single dropdown `Edit` button which had a menu of two options: `Edit` and `Delete`.

The new design splits them up into two discrete buttons, hence we create two button components.

**Improvements**:

- By extending `ChakraButtonProps`, we are able to pass the chakra-default props into our own buttons. This allow us to pass down useful props like `variant`, `color`, etc from the top-level component. 

**Bug Fixes**:

- Fixes #1517 , where the `Post` button is too close to the `AccordionMenu`

## Before & After Screenshots
**Desktop**:
BEFORE
Screenshot 2022-04-12 at 5 07 39 PM
AFTER
Screenshot 2022-04-12 at 5 04 20 PM

**Mobile**:
BEFORE
image

AFTER
Screenshot 2022-04-12 at 5 06 12 PM

## Testing
Changes are live on staging

PR #1575, opened 2022-04-18
Release 2022-04-18

LoneRifle:
This release contains the following:
- #1350
- #1523 
- #1564 
- #1554 
- #1553 
- #1549 
- #1548


christopherlim98 (approved):
lgtm! style changes looks good on staging

One small unrelated UX issue is that scroll to top does not work when changing between topics. Filing an issue here #1576

zxt-tzx (approved):
lgtm

PR #1616, opened 2022-04-27
fix(post): respond with 404 on post not found

LoneRifle:
## Problem and Solution
Minimise spurious http 500 errors by returning the correct status code
if post is not found

- reconfigure tsc to emit es6 so that `instanceof` is meaningful
- if caught error is PublicPostNotFoundError, respond with that error's
  message and status code, not the generic one
- provide unit test coverage to account for 404


christopherlim98 (approved):
lgtm!

